PR #7, opened 2024-01-15
Feature flag support

linglingye001:
- Feature flags can be sourced to configMap
- Not support `default` and `properties` configMap data type when `featureFlags` filed set
- Refresh of feature flag is time-based, not influenced by sentinel key change
- Separator will be ignored for feature flags(consistent with CLI)

 API 
```
apiVersion: azconfig.io/v1
kind: AzureAppConfigurationProvider
metadata:
  name: appconfigurationprovider-sample
spec:
  endpoint: <app-configuration-store-endpoint>
  target:
    configMapName: configmap-created-by-appconfig-provider        
  configuration:
    selectors:
      - keyFilter: app1*
        labelFilter: common
  featureFlag:
    selectors:        
      - keyFilter: 
        labelFilter:
    refresh:   
      enabled:    
      interval:  
```


internal/loader/configuration_setting_loader.go
```diff
@@ -471,37 +579,52 @@ func getSecret(ctx context.Context, namespacedSecretName types.NamespacedName) (
 	return secretObject, nil
 }
 
-func getFilters(acpSpec acpv1.AzureAppConfigurationProviderSpec) []acpv1.KeyValueSelector {
-	var filters []acpv1.KeyValueSelector
+func getFilters(acpSpec acpv1.AzureAppConfigurationProviderSpec) ([]acpv1.KeyValueSelector, []acpv1.KeyValueSelector) {
```
RichardChen820, line 582: Don't use one method to get both `KeyValueSelector` and `KeyValueSelector`. Split to two methods

api/v1/azureappconfigurationprovider_types.go
```diff
@@ -69,6 +71,12 @@ type AzureAppConfigurationKeyValueOptions struct {
...
 }
 
+// AzureAppConfigurationFeatureFlagOptions defines the options of fetching feature flags from AppConfiguration.
+type AzureAppConfigurationFeatureFlagOptions struct {
+	Selectors []KeyValueSelector `json:"selectors,omitempty"`
+	Refresh   *RefreshSettings   `json:"refresh,omitempty"`
+}
+
 // KeyValueSelector defines the filters when fetching the data from Azure AppConfiguration
 type KeyValueSelector struct {
```
RichardChen820, line 81: How about rename it to `KeyLabelSelector` ?  Since it's not only for selecting key-values, but also for select feature flags

internal/loader/configuration_setting_loader.go
```diff
@@ -109,21 +122,92 @@ func NewConfigurationSettingLoader(ctx context.Context, provider acpv1.AzureAppC
 	}, nil
 }
 
-func (csp *ConfigurationSettingLoader) CreateKeyValueSettings(ctx context.Context, resolveSecretReference ResolveSecretReference) (*TargetKeyValueSettings, error) {
+func (csl *ConfigurationSettingLoader) CreateTargetSettings(ctx context.Context, resolveSecretReference ResolveSecretReference) (*TargetKeyValueSettings, error) {
+	unformattedSettings, err := csl.CreateKeyValueSettings(ctx, resolveSecretReference)
```
RichardChen820, line 126: How about call it `untypedSettings` to be symmetric to `typedSettings`?

RichardChen820 (changes requested):

internal/loader/configuration_setting_loader.go
RichardChen820: After revisiting this, I think `rawSettings` is better, thoughts?

internal/loader/typed_setting.go
```diff
@@ -15,23 +15,23 @@ import (
 	"k8s.io/klog/v2"
 )
 
-func createTypedSettings(settings map[string]*string, isJsonContentTypeMap map[string]bool, dataOptions *acpv1.ConfigMapDataOptions) (map[string]string, error) {
-	if settings == nil {
+func createTypedSettings(unformattedSettings *UnformattedSettings, dataOptions *acpv1.ConfigMapDataOptions) (map[string]string, error) {
+	if unformattedSettings.KeyValueSettings == nil {
```
RichardChen820, line 19: Is the possiable that the KeyValueSettings and IsJsonContentTypeMap is nil?

internal/loader/typed_setting.go
```diff
@@ -118,3 +116,80 @@ func isJsonContentType(contentType *string) bool {
...
+	if featureFlagSettings == nil {
+		return nil, nil
+	}
+
+	switch featureFlagSettings.(type) {
+	// feature flag settings from existing configMap
+	case map[string]interface{}:
+		return featureFlagSettings, nil
+	// feature flag settings from appconfig store
+	case []*string:
```
RichardChen820, line 176: I can't catch up why we need to handle this type. feature flag settings from appconfig store should be also `map[string]interface{}` type

internal/loader/configuration_setting_loader.go
```diff
@@ -43,13 +43,23 @@ type TargetKeyValueSettings struct {
 	KeyVaultReferencesToCache map[string]KeyVaultSecretUriSegment
 }
 
+type UnformattedSettings struct {
+	KeyValueSettings          map[string]*string
+	IsJsonContentTypeMap      map[string]bool
+	FeatureFlagSettings       interface{}
```
RichardChen820, line 49: I would expect this is `map[string]interface{}` type

internal/loader/typed_setting.go
```diff
@@ -118,3 +116,80 @@ func isJsonContentType(contentType *string) bool {
 	matched, _ := regexp.MatchString("^application\\/(?:[^\\/]+\\+)?json(;.*)?$", contentTypeStr)
 	return matched
 }
+
+func UnmarshalConfigMap(existingConfigMapSetting *map[string]string, dataOptions *acpv1.ConfigMapDataOptions) (interface{}, map[string]interface{}, error) {
+	allSettings := make(map[string]interface{})
+	existingConfigMapData := (*existingConfigMapSetting)[dataOptions.Key]
+	var typedFeatureFlags interface{}
+	var err error
+
```
RichardChen820, line 125: Please comments that type is not possible to be `default` or `properties` here

internal/loader/typed_setting.go
```diff
@@ -118,3 +116,80 @@ func isJsonContentType(contentType *string) bool {
 	matched, _ := regexp.MatchString("^application\\/(?:[^\\/]+\\+)?json(;.*)?$", contentTypeStr)
 	return matched
 }
+
+func UnmarshalConfigMap(existingConfigMapSetting *map[string]string, dataOptions *acpv1.ConfigMapDataOptions) (interface{}, map[string]interface{}, error) {
+	allSettings := make(map[string]interface{})
+	existingConfigMapData := (*existingConfigMapSetting)[dataOptions.Key]
```
RichardChen820, line 122: What if the existingConfigMapSetting is `default` type?

internal/loader/typed_setting.go
```diff
@@ -118,3 +116,80 @@ func isJsonContentType(contentType *string) bool {
...
+	var typedFeatureFlags interface{}
+	var err error
+
+	switch dataOptions.Type {
+	case acpv1.Yaml:
```
RichardChen820, line 127: Please comments that type has no possibility of being `default` or `properties` here

internal/controller/processor.go
```diff
@@ -38,16 +38,22 @@ type RefreshOptions struct {
...
 }
 
-func (processor *AppConfigurationProviderProcessor) PopulateSettings(existingSecret *corev1.Secret) error {
+func (processor *AppConfigurationProviderProcessor) PopulateSettings(existingConfigMap *corev1.ConfigMap, existingSecret *corev1.Secret) error {
 	if err := processor.ProcessFullReconciliation(); err != nil {
 		return err
 	}
 
-	if err := processor.ProcessKeyValueRefresh(); err != nil {
+	if err := processor.ProcessFeatureFlagRefresh(existingConfigMap); err != nil {
```
RichardChen820, line 52: Seems if both `FeatureFlagRefresh` and `KeyValueRefresh` are triggered in same reconciliation. There would be confliction.

internal/loader/typed_setting.go
linglingye001: `UnmarshalConfigMap` will return typed feature flag settings. When retrieving from appconfig store, feature flag settings are JSON-encoded data which need to call `json.Unmarshal` and store the values in `interface{}` for subsequent `json/yaml` format configMap data generation.

internal/controller/processor.go
linglingye001: When they are triggered in same reconciliation, `processor` will sequentially update parts of the feature flag and key-value. In their respective update processes, the other content will remain unchanged. What does conflict specifically refer to？

internal/loader/configuration_setting_loader.go
```diff
@@ -197,15 +282,47 @@ func (csp *ConfigurationSettingLoader) CreateKeyValueSettings(ctx context.Contex
 	}
 
 end:
-	typedSetting, err := createTypedSettings(configMapSettings, isJsonContentTypeMap, csp.Spec.Target.ConfigMapData)
-	if err != nil {
-		return nil, err
+	return rawSettings, nil
+}
+
+func (csl *ConfigurationSettingLoader) getFeatureFlagSettings(ctx context.Context) (map[string]interface{}, error) {
```
RichardChen820, line 288: Shouldn't the return type be `map[string][]interface{}`?

internal/controller/utils.go
```diff
@@ -18,6 +18,7 @@ import (
 const (
 	MinimalSentinelBasedRefreshInterval time.Duration = time.Second
 	MinimalSecretRefreshInterval        time.Duration = time.Minute
+	MinimalFeatureFlagRefreshInterval   time.Duration = time.Second
```
RichardChen820, line 21: 1 second maybe too short for the MinimalFeatureFlagRefreshInterval

internal/controller/utils.go
```diff
@@ -59,6 +60,29 @@ func verifyObject(spec acpv1.AzureAppConfigurationProviderSpec) error {
 		}
 	}
 
+	if spec.FeatureFlag != nil {
+		if spec.Target.ConfigMapData == nil || spec.Target.ConfigMapData.Type == acpv1.Default || spec.Target.ConfigMapData.Type == acpv1.Properties {
+			return loader.NewArgumentError("spec.target.configMapData", fmt.Errorf("target.configMapData.type must be json or yaml when FeatureFlag is set"))
+		}
+
```
RichardChen820, line 67: Should validate the length of selector can't be 0

internal/loader/configuration_setting_loader.go
linglingye001: `map[string]interface{}` is used to maintain consistency with the type returned by `UnmarshalConfigMap`

RichardChen820 (approved):

internal/controller/utils.go
```diff
@@ -59,6 +60,33 @@ func verifyObject(spec acpv1.AzureAppConfigurationProviderSpec) error {
...
+		if len(spec.FeatureFlag.Selectors) == 0 {
+			return loader.NewArgumentError("spec.featureFlag.selectors", fmt.Errorf("featureFlag.selectors must be specified when FeatureFlag is set"))
+		}
+
+		// Check if feature flag label filters are valid
+		if len(spec.FeatureFlag.Selectors) > 0 {
```
RichardChen820, line 73: This judgement is not necessary

api/v1/azureappconfigurationprovider_types.go
```diff
@@ -64,13 +66,19 @@ type ConfigurationGenerationParameters struct {
...
-// KeyValueSelector defines the filters when fetching the data from Azure AppConfiguration
-type KeyValueSelector struct {
+// AzureAppConfigurationFeatureFlagOptions defines the options of fetching feature flags from AppConfiguration.
+type AzureAppConfigurationFeatureFlagOptions struct {
+	Selectors []KeyLabelSelector `json:"selectors,omitempty"`
+	Refresh   *RefreshSettings   `json:"refresh,omitempty"`
+}
+
+// KeyLabelSelector defines the filters when fetching the data from Azure AppConfiguration
+type KeyLabelSelector struct {
```
RichardChen820, line 81: We will add `snapshotName` in the selector in the near future, so `KeyLabelSelector` is not adaptive enough. Do you mind rename this to `Selector`?

internal/controller/processor.go
```diff
@@ -75,7 +81,44 @@ func (processor *AppConfigurationProviderProcessor) ProcessFullReconciliation()
...
+		reconcileState.NextFeatureFlagRefreshReconcileTime = nextFeatureFlagRefreshReconcileTime
+		return nil
+	}
+
+	if !currentTime.After(reconcileState.NextFeatureFlagRefreshReconcileTime.Time) {
+		return nil
+	}
+
+	processor.RefreshOptions.featureFlagRefreshNeeded = true
+	partiallyUpdatedSettings, err := (*processor.Retriever).RefreshFeatureFlagSettings(processor.Context, &existingConfigMap.Data)
```
RichardChen820, line 109: Do you mind rename `partiallyUpdatedSettings` to `featureFlagRefreshedSettings`?

internal/controller/processor.go
```diff
@@ -107,20 +150,22 @@ func (processor *AppConfigurationProviderProcessor) ProcessKeyValueRefresh() err
...
 		return nil
 	}
-
 	// Get the latest key value settings
-	updatedSettings, err := (*processor.Retriever).CreateKeyValueSettings(processor.Context, processor.ResolveSecretReference)
+	existingConfigMapSettings := &existingConfigMap.Data
+	if processor.Settings.ConfigMapSettings != nil {
+		existingConfigMapSettings = &processor.Settings.ConfigMapSettings
+	}
+	partiallyUpdatedSettings, err := (*processor.Retriever).RefreshKeyValueSettings(processor.Context, existingConfigMapSettings, processor.ResolveSecretReference)
```
RichardChen820, line 158: Do you mind rename `partiallyUpdatedSettings` to `keyValueRefreshedSettings`?

linglingye001: Updated

internal/loader/typed_setting.go
```diff
@@ -118,3 +104,51 @@ func isJsonContentType(contentType *string) bool {
...
+
+	if featureFlagSection, ok := allSettings[FeatureManagementSectionName].(map[string]interface{}); ok {
+		typedFeatureFlags = featureFlagSection
+		delete(allSettings, FeatureManagementSectionName)
+	}
+
+	return typedFeatureFlags, allSettings, nil
+}
+
+func createJsonYamlTypeData(settings map[string]interface{}, dataOptions *acpv1.ConfigMapDataOptions) (string, error) {
```
RichardChen820, line 137: Do you mind rename this to `marshalJsonYaml` ?

internal/loader/typed_setting.go
```diff
@@ -118,3 +104,51 @@ func isJsonContentType(contentType *string) bool {
 	matched, _ := regexp.MatchString("^application\\/(?:[^\\/]+\\+)?json(;.*)?$", contentTypeStr)
 	return matched
 }
+
+// UnmarshalConfigMap is used for scenarios related to feature flag refresh.
+// Currently, only supports feature flag settings in JSON/YAML formats.
+func UnmarshalConfigMap(existingConfigMapSetting *map[string]string, dataOptions *acpv1.ConfigMapDataOptions) (map[string]interface{}, map[string]interface{}, error) {
```
RichardChen820, line 110: Why this is Pascal type but `createJsonYamlTypeData` is Camel style?

internal/loader/typed_setting.go
```diff
@@ -118,3 +104,51 @@ func isJsonContentType(contentType *string) bool {
...
+	allSettings := make(map[string]interface{})
+	existingConfigMapData := (*existingConfigMapSetting)[dataOptions.Key]
```
RichardChen820, line 112: This method `UnmarshalConfigMap` should only be called during refreshing process, so I know there should be no reason that `(*existingConfigMapSetting)[dataOptions.Key]` does not exists, hence we'd better handle it by throwing an exception if `dataOptions.Key` does not exist, instead of a silent error

internal/loader/typed_setting.go
```diff
@@ -118,3 +104,51 @@ func isJsonContentType(contentType *string) bool {
...
+	var typedFeatureFlags map[string]interface{}
+	var err error
+
+	switch dataOptions.Type {
```
RichardChen820, line 116: We should safe guard that the type should be either `json` or `yaml`. Could we throw exception here when data type is not `json` or `yaml`?

internal/loader/typed_setting.go
```diff
@@ -118,3 +104,51 @@ func isJsonContentType(contentType *string) bool {
...
+			return nil, nil, err
+		}
+	}
+
+	if featureFlagSection, ok := allSettings[FeatureManagementSectionName].(map[string]interface{}); ok {
+		typedFeatureFlags = featureFlagSection
+		delete(allSettings, FeatureManagementSectionName)
+	}
+
+	return typedFeatureFlags, allSettings, nil
```
RichardChen820, line 134: it's not really allSettings are returned, do you mind rename `allSettings` to `keyValueSettings`?

internal/controller/utils.go
```diff
@@ -59,6 +60,31 @@ func verifyObject(spec acpv1.AzureAppConfigurationProviderSpec) error {
 		}
 	}
 
+	if spec.FeatureFlag != nil {
+		if spec.Target.ConfigMapData == nil || spec.Target.ConfigMapData.Type == acpv1.Default || spec.Target.ConfigMapData.Type == acpv1.Properties {
+			return loader.NewArgumentError("spec.target.configMapData", fmt.Errorf("target.configMapData.type must be json or yaml when FeatureFlag is set"))
```
RichardChen820, line 65: ArgumentError already indicate the setting section is `spec.target.configMapData`. So the message should be `configMap data type must be json or yaml when FeatureFlag is set.`

internal/loader/typed_setting.go
```diff
@@ -118,3 +104,57 @@ func isJsonContentType(contentType *string) bool {
...
+
+// Used for scenarios related to feature flag refresh.
+// Currently, only supports feature flag settings in JSON/YAML formats.
+func unmarshalConfigMap(existingConfigMapSetting *map[string]string, dataOptions *acpv1.ConfigMapDataOptions) (map[string]interface{}, map[string]interface{}, error) {
+	keyValueSettings := make(map[string]interface{})
+	var typedFeatureFlags map[string]interface{}
+
+	if existingConfigMapData, ok := (*existingConfigMapSetting)[dataOptions.Key]; !ok {
+		return nil, nil, fmt.Errorf("Failed to get data for key '%s' from configMap", dataOptions.Key)
+	} else {
```
RichardChen820, line 116: This `else` is not necessary

internal/loader/typed_setting.go
```diff
@@ -118,3 +104,57 @@ func isJsonContentType(contentType *string) bool {
...
+			if err != nil {
+				return nil, nil, err
+			}
+		default:
+			return nil, nil, fmt.Errorf("Unsupported data type '%s', only supports json/yaml for now", dataOptions.Type)
+		}
+
+		if featureFlagSection, ok := keyValueSettings[FeatureManagementSectionName].(map[string]interface{}); !ok {
+			return nil, nil, fmt.Errorf("Failed to get feature flags from configMap data '%s'", existingConfigMapData)
+		} else {
```
RichardChen820, line 134: not necessary `else`

PR #8, opened 2024-01-24
Support setting tolerations, nodeSelector, and affinity during helm install

linglingye001:


deploy/templates/autoscaling.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: autoscaling/v2
```
RichardChen820, line 1: Shouldn't he HPA only be deployed while the autoscaling is enabled?

deploy/templates/autoscaling.yaml
```diff
@@ -0,0 +1,26 @@
+{{- if .Values.autoscaling.enabled }}
+apiVersion: autoscaling/v2
+kind: HorizontalPodAutoscaler
+metadata:
+  name: az-appconfig-k8s-provider-hpa
```
RichardChen820, line 5: use the full name variable

deploy/parameter/helm-values.yaml
```diff
@@ -41,18 +41,31 @@ securityContext:
 
 resources: 
   limits:
-    cpu: 500m
+    cpu: 100m
     memory: 128Mi
   requests:
-    cpu: 10m
+    cpu: 20m
```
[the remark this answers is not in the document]
linglingye001: image

deploy/parameter/helm-values.yaml
```diff
@@ -41,18 +41,31 @@ securityContext:
 
 resources: 
   limits:
-    cpu: 500m
+    cpu: 100m
     memory: 128Mi
```
RichardChen820, line 45: Seems double the request memory is too conservative. How about using 256Mi as a limit?

linglingye001: Updated

RichardChen820 (approved):

PR #10, opened 2024-01-30
Support generating cert based TLS type secret

RichardChen820:
Support generating cert-based TLS type secret


internal/loader/configuration_setting_loader.go
```diff
@@ -230,17 +245,26 @@ func (csl *ConfigurationSettingLoader) CreateKeyValueSettings(ctx context.Contex
...
+				case SecretReferenceContentType:
 					if setting.Value == nil {
 						return nil, fmt.Errorf("The value of Key Vault reference '%s' is null", *setting.Key)
 					}
-					if csl.Spec.Secret == nil {
+
+					var secretType corev1.SecretType = corev1.SecretTypeOpaque
+					var err error
+					if secretTypeTag, ok := setting.Tags[PreservedSecretTypeTag]; ok {
+						secretType, err = parseSecretType(secretTypeTag)
```
linglingye001, line 257: `parseSecretType` looks like only supporting tls type? Can user set opaque type through tag?

linglingye001, line 257: `parseSecretType` looks like only supporting tls type? Can user set opaque type through tag?

internal/loader/configuration_setting_loader.go
```diff
@@ -230,17 +245,26 @@ func (csl *ConfigurationSettingLoader) CreateKeyValueSettings(ctx context.Contex
...
+						if err != nil {
+							return nil, err
+						}
+					} else if csl.Spec.Secret == nil {
```
linglingye001, line 261: Does the `else if` statement mean `Spec.Secret` can be `nil` when secret type is tls? If so, it may miss credentials when create `SecretReferenceResolver`.

internal/loader/configuration_setting_loader.go
RichardChen820: No, current I don't want to support it, if someone want an opaque secret, just don't set any tag is fine

internal/controller/appconfigurationprovider_controller.go
```diff
@@ -129,37 +129,64 @@ func (reconciler *AzureAppConfigurationProviderReconciler) Reconcile(ctx context
...
-
-	/* Initialize the ReconcileState for the provider*/
-	if reconciler.ProvidersReconcileState[req.NamespacedName] == nil {
+		existingSecrets[provider.Spec.Secret.Target.SecretName] = existingSecret
+	}
+
+	if reconciler.ProvidersReconcileState[req.NamespacedName] != nil {
+		for name := range reconciler.ProvidersReconcileState[req.NamespacedName].ExistingSecretReferences {
+			if _, ok := existingSecrets[name]; !ok {
+				existingSecret = corev1.Secret{
```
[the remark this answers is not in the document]
RichardChen820: After revisiting this code snippet, seems that situation won't happen. `existingSecrets` should always be the current one.

internal/controller/appconfigurationprovider_controller.go
```diff
@@ -129,37 +129,64 @@ func (reconciler *AzureAppConfigurationProviderReconciler) Reconcile(ctx context
...
+			ExistingSecretReferences: make(map[string]*loader.TargetSecretReference),
 		}
 	}
+
 	// Reset the resource version if the configmap or secret was unexpected deleted
 	if existingConfigMap.Name == "" {
 		reconciler.ProvidersReconcileState[req.NamespacedName].ConfigMapResourceVersion = nil
 	}
+
 	if provider.Spec.Secret == nil || existingSecret.Name == "" {
```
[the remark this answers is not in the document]
RichardChen820: It's possible, when a yaml is initialized with `Secret` setting, that's the situation of `spec.secret != nil,` but `existingSecret.Name == ""`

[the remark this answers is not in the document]
RichardChen820: Good catch! It should be updated accordingly

api/v1/azureappconfigurationprovider_types.go
```diff
@@ -141,8 +141,8 @@ type ManagedIdentityReferenceParameters struct {
 	Key string `json:"key"`
 }
 
-// AzureKeyVaultReference defines the authentication type used to Azure KeyVault resolve KeyVaultReference
-type AzureKeyVaultReference struct {
+// SecretReference defines the authentication type used to Azure KeyVault resolve KeyVaultReference
```
juniwang, line 144: The description can be updated to match the latest usage. It contains more supported properties than auth.

internal/controller/appconfigurationprovider_controller.go
```diff
@@ -379,28 +467,41 @@ func newProviderStatus(phase acpv1.AppConfigurationSyncPhase, message string, sy
...
+	if len(reconciler.ProvidersReconcileState[namespacedName].ExistingSecretReferences) == 0 {
 		return true
 	}
 
+	for name, secret := range existingSecrets {
+		if reconciler.ProvidersReconcileState[namespacedName].ExistingSecretReferences[name].SecretResourceVersion != secret.ResourceVersion {
+			return true
+		}
+	}
+
```
juniwang, line 504: Is it possible that the loop of updating secret in `createOrUpdateSecrets` breaks early? ending up with a state that `len(reconciler.ProvidersReconcileState[namespacedName].ExistingSecretReferences) == 0` is true and `ExistingSecretReferences[name].SecretResourceVersion == secret.ResourceVersion` is true for all existing secrets, but the ir lengths don't match the length of `settings.SecretSettings` hence needs reconcile.

internal/loader/configuration_setting_loader.go
```diff
@@ -679,3 +778,108 @@ func createSecretClients(ctx context.Context, acp acpv1.AzureAppConfigurationPro
...
+
+func parseSecretType(secretType string) (corev1.SecretType, error) {
+	secretTypeMap := map[string]corev1.SecretType{
+		"opaque":            corev1.SecretTypeOpaque,
+		"kubernetes.io/tls": corev1.SecretTypeTLS,
+	}
+
+	if parsedType, ok := secretTypeMap[secretType]; ok {
+		if parsedType != corev1.SecretTypeTLS {
+			return "", fmt.Errorf("secret type %q is not supported", secretType)
```
juniwang, line 790: So we don't allow customers to specify secretType `opaque` via the preserved tag?

internal/loader/configuration_setting_loader.go
RichardChen820: Good catch, updated

api/v1/azureappconfigurationprovider_types.go
RichardChen820: Updated

internal/controller/appconfigurationprovider_controller.go
RichardChen820: Updated. And moved this `shouldReconcile()` function to `processor` to make the detailed logic transparency to the controller.

internal/loader/configuration_setting_loader.go
RichardChen820: The default Secret being generated is `opaque`. There's no reasonable circumstance to specify `opaque` as preserved tag

internal/loader/configuration_setting_loader.go
```diff
@@ -349,33 +385,78 @@ func (csl *ConfigurationSettingLoader) CheckAndRefreshSentinels(ctx context.Cont
...
+				if resolvedSecret.Kid != nil && len(*resolvedSecret.Kid) > 0 {
+					switch *resolvedSecret.ContentType {
+					case CertTypePfx:
+						resolvedSecretReferences[name].Data[TlsKey], resolvedSecretReferences[name].Data[TlsCrt], err = decodePkcs12(*resolvedSecret.Value)
+					case CertTypePem:
+						resolvedSecretReferences[name].Data[TlsKey], resolvedSecretReferences[name].Data[TlsCrt], err = decodePem(*resolvedSecret.Value)
+					default:
+						err = fmt.Errorf("failed to get certificate, unknown content type '%s'", *resolvedSecret.ContentType)
+					}
+				} else {
```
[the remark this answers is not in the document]
RichardChen820: I was looking into this part as well, no need to have the `if..else..` block, just need to check the contentType and take relevant action.

internal/controller/processor.go
```diff
@@ -172,62 +171,120 @@ func (processor *AppConfigurationProviderProcessor) ProcessKeyValueRefresh(exist
...
+		*processor.ReconciliationState.ConfigMapResourceVersion != existingConfigMap.ResourceVersion {
+		// If the ConfigMap is removed or updated, we need to reconcile anyway
+		return true
+	}
+
+	if processor.Provider.Spec.Secret == nil {
+		return false
+	}
+
+	if len(processor.ReconciliationState.ExistingSecretReferences) == 0 ||
```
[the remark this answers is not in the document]
RichardChen820: Consider this scenario:
In beginning,  no secret-related setting being specified in the yaml. Then user updates the yaml to add Secret. In this case, `len(processor.ReconciliationState.ExistingSecretReferences) == 0` and need to reconcile

internal/controller/processor.go
```diff
@@ -25,83 +24,83 @@ type AppConfigurationProviderProcessor struct {
 	Provider               *acpv1.AzureAppConfigurationProvider
 	Settings               *loader.TargetKeyValueSettings
 	ShouldReconcile        bool
-	ReconciliationState    map[types.NamespacedName]*ReconciliationState
+	ReconciliationState    *ReconciliationState
 	CurrentTime            metav1.Time
-	NamespacedName         types.NamespacedName
 	RefreshOptions         *RefreshOptions
-	ResolveSecretReference loader.ResolveSecretReference
+	ResolveSecretReference loader.SecretReferenceResolver
```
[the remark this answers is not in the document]
RichardChen820: Updated

internal/controller/processor.go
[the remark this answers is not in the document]
linglingye001: If user updates the yaml, it can be returned early in line245. Do es this scenario happen when user updates the yaml after seeing error log? Also there's a scenario when user specified `secret` section in yaml but no key vault references in store, which may also make `len(ExistingSecretReferences)==0`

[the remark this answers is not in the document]
RichardChen820: >> Also there's a scenario when user specified secret section in yaml but no key vault references in store, which may also make len(ExistingSecretReferences)==0

No, as long as user specify the secret section, at least one secret will be generate even it's empty.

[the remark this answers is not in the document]
linglingye001: I see, thanks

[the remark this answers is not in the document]
RichardChen820: > If user updates the yaml, it can be returned early in line245.
Right, but `len(processor.ReconciliationState.ExistingSecretReferences) == 0` is an unexpected scenario, once it happens means something was wrong(may be just because the reconciliation never succeeded)

linglingye001 (approved):

PR #17, opened 2024-03-04
Add Snapshot selector support

RichardChen820:
Add Snapshot selector support


api/v1/azureappconfigurationprovider_types.go
```diff
@@ -79,8 +79,9 @@ type AzureAppConfigurationFeatureFlagOptions struct {
 
 // KeyLabelSelector defines the filters when fetching the data from Azure AppConfiguration
 type Selector struct {
-	KeyFilter   string  `json:"keyFilter"`
-	LabelFilter *string `json:"labelFilter,omitempty"`
+	KeyFilter    *string `json:"keyFilter,omitempty"`
+	LabelFilter  *string `json:"labelFilter,omitempty"`
+	SnapshotName *string `json:"snapshotName,omitempty"`
```
juniwang, line 84: why is this not named `xxxFilter`? sounds inconsistent

internal/controller/utils.go
```diff
@@ -71,21 +71,21 @@ func verifyObject(spec acpv1.AzureAppConfigurationProviderSpec) error {
 
 		// Check if feature flag label filters are valid
 		for i := range spec.FeatureFlag.Selectors {
-			if spec.FeatureFlag.Selectors[i].LabelFilter != nil && hasNonEscapedValueInLabel(*spec.FeatureFlag.Selectors[i].LabelFilter) {
-				return loader.NewArgumentError("spec.featureFlag.selectors", fmt.Errorf("non-escaped reserved wildcard character '*' and multiple labels separator ',' are not supported in label filters"))
+			err = verifySelectorObject(spec.FeatureFlag.Selectors[i])
```
juniwang, line 74: FeatureFlag is verifying input the same way asa Configuration.Selectors. Does this mean FeatureFlag supports Snapshot filter now?

internal/loader/configuration_setting_loader.go
```diff
@@ -535,26 +538,48 @@ func getConfigurationSettings(
...
+		} else {
+			snapshot, err := client.GetSnapshot(ctx, *filter.SnapshotName, nil)
 			if err != nil {
 				e <- err
-			} else if len(page.Settings) > 0 {
-				c <- page.Settings
 			}
-		}
 
+			if *snapshot.CompositionType != azappconfig.CompositionTypeKey {
```
juniwang, line 566: why only accept 'key' type and not allowing key+lable?

api/v1/azureappconfigurationprovider_types.go
RichardChen820: This is discussed in the design, `snapshotFilter` sounds not reasonable.

internal/loader/configuration_setting_loader.go
RichardChen820: This is by design, it's consistent with other providers. support key+label type would cause ambiguity problem, can't which label would be populated is not definitive

internal/controller/utils.go
RichardChen820: Yes, it will filter out all the feature flag type key-values in specified snapshot

internal/loader/configuration_setting_loader.go
```diff
@@ -535,26 +538,48 @@ func getConfigurationSettings(
...
+			for pager.More() {
+				page, err := pager.NextPage(ctx)
+				if err != nil {
+					e <- err
+				} else if len(page.Settings) > 0 {
+					c <- page.Settings
+				}
+			}
+		} else {
+			snapshot, err := client.GetSnapshot(ctx, *filter.SnapshotName, nil)
```
linglingye001, line 561: Can we add snapshot status check since it can be archived? And what's expected behavior if  the snapshot expire after some interval?

internal/loader/configuration_setting_loader.go
```diff
@@ -693,7 +718,10 @@ func getFeatureFlagFilters(acpSpec acpv1.AzureAppConfigurationProviderSpec) []ac
 	if acpSpec.FeatureFlag != nil {
 		featureFlagFilters = deduplicateFilters(acpSpec.FeatureFlag.Selectors)
 		for i := 0; i < len(featureFlagFilters); i++ {
-			featureFlagFilters[i].KeyFilter = FeatureFlagKeyPrefix + featureFlagFilters[i].KeyFilter
+			if featureFlagFilters[i].KeyFilter != nil {
```
linglingye001, line 721: If user set snapshot in feature flag section, what  is expected behavior? Ignoring all `featureFlag.selector.snapshot` ?

internal/loader/configuration_setting_loader.go
RichardChen820: Archived snapshots are still available to be used.

internal/loader/configuration_setting_loader.go
```diff
@@ -337,12 +337,15 @@ func (csl *ConfigurationSettingLoader) getFeatureFlagSettings(ctx context.Contex
 				goto end
 			} else {
 				for _, setting := range featureFlagSettings {
-					var out interface{}
-					err := json.Unmarshal([]byte(*setting.Value), &out)
-					if err != nil {
-						return nil, fmt.Errorf("Failed to unmarshal feature flag settings: %s", err.Error())
+					if setting.ContentType != nil &&
```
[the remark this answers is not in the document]
RichardChen820: Same with what we filtered out by key/lable filters

internal/controller/appconfigurationprovider_controller_test.go
```diff
@@ -396,6 +396,7 @@ var _ = Describe("AppConfiguationProvider controller", func() {
 			ctx := context.Background()
 			providerName := "test-appconfigurationprovider-8"
 			configMapName := "file-style-configmap-to-be-created-2"
+			wildcard := "*"
```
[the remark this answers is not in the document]
RichardChen820: We allow it , but we don't treat it as wildcard, it just a normal character

internal/loader/configuration_setting_loader.go
RichardChen820: No, we go ahead to use it, get the snapshot, and retrieve the feature flags settings from it.

linglingye001 (approved):

Issue #32, opened 2024-04-22
Cannot update existing secrets from the 1.3.0 version


antoineozenne-at-leocare:
I updated from version `1.2.0` to `1.3.0`, but I have the following error for all my secrets (with a verbosity to 5):

```console
I0422 13:02:14.590953       1 appconfigurationprovider_controller.go:396] No secret settings are fetched from Azure AppConfiguration
W0422 13:02:14.672973       1 appconfigurationprovider_controller.go:316] Fail to update the target ConfigMap or Secret of AzureAppConfigurationProvider 'foo' in 'bar' namespace: secrets "foo" is forbidden: User "system:serviceaccount:azappconfig-system:az-appconfig-k8s-provider" cannot delete resource "secrets" in API group "" in the namespace "bar": Azure does not have opinion for this user.
```

I noticed that the `existingSecrets` variable is no longer a pointer in the version `1.3.0` in https://github.com/Azure/AppConfiguration-KubernetesProvider/blob/26d01225215c793585ccfcfaaa0708bddfb5eb08/internal/controller/appconfigurationprovider_controller.go#L239

I think this could be the cause of my problem. I did not change anything in my `AzureAppConfigurationProvider` resources which use `secret.auth.workloadIdentity.managedIdentityClientIdReference`.

RichardChen820:
Regression in v1.3.0, while no secret is selected by the selectors, it fails to create an empty target Secret.

RichardChen820:
@antoineozenne-at-leocare The regression should be fixed in v1.3.1, could you give it another try?

antoineozenne-at-leocare:
> @antoineozenne-at-leocare The regression should be fixed in v1.3.1, could you give it another try?

Now it works, thank you @RichardChen820!

PR #34, opened 2024-04-23
Fix the secret creation and update issue

RichardChen820:
Fix the secret creation and update issue.https://github.com/Azure/AppConfiguration-KubernetesProvider/issues/32


internal/controller/appconfigurationprovider_controller.go
```diff
@@ -250,6 +250,22 @@ func (reconciler *AzureAppConfigurationProviderReconciler) Reconcile(ctx context
 
 	/* Create secret when there are secret settings */
 	if processor.RefreshOptions.SecretSettingPopulated {
+		// Verify the existence of the secret which is not owned by the current provider
+		for name := range processor.Settings.SecretSettings {
+			if _, ok := existingSecrets[name]; !ok {
```
juniwang, line 255: telling from https://github.com/Azure/AppConfiguration-KubernetesProvider/blob/main/internal/controller/processor.go#L229: 
```
processor.Settings.SecretSettings = existingSecrets
```
Does this mean `processor.Settings.SecretSettings` and `existingSecrets` reference to the same object?

internal/controller/appconfigurationprovider_controller.go
```diff
@@ -250,6 +250,22 @@ func (reconciler *AzureAppConfigurationProviderReconciler) Reconcile(ctx context
 
 	/* Create secret when there are secret settings */
 	if processor.RefreshOptions.SecretSettingPopulated {
+		// Verify the existence of the secret which is not owned by the current provider
+		for name := range processor.Settings.SecretSettings {
```
juniwang, line 254: aslo I don't understand the code at https://github.com/Azure/AppConfiguration-KubernetesProvider/blob/main/internal/controller/processor.go#L223C2-L228C3.

```
	for secretName, targetSecret := range resolvedSecretData {
		existingSecret, ok := existingSecrets[secretName]
		if ok {
			maps.Copy(existingSecret.Data, targetSecret.Data)
		}
	}
```

RichardChen820: That's because we only resolve the non-versioned secret, so we need to merge the new values into the existing secret to update the non-versioned secrets

internal/controller/appconfigurationprovider_controller.go
RichardChen820: Correct, it won't be a problem

juniwang: if they reference to the same object, the check here seems to be unnecessary.  when will `if _, ok := existingSecrets[name];` returns a false value?

internal/controller/appconfigurationprovider_controller.go
juniwang: after checking go lang doc, I know why it confused me. `maps.Copy` actually copies data from the second parameter to the first, not the other way around. The parameter `targetSecret` makes me think the second is the destination. Can we rename it to `resolvedSecret`?

internal/controller/appconfigurationprovider_controller.go
RichardChen820: `existingSecrets` is caculated everytime of reconciliation. The `resolvedSecrets` is resolved from `secretReferencesToSolve`.  In normal situations, the keys of them should be the same, we don't really need this check. But we can't 100% sure that the `existingSecrets[secretName]` exists because the customer may accidently remove that secret, although we have correction mechanism to re-generate the removed secret, but there's may has a timing that the key of these two object is not match. Add this judge to safe guard not crashing the process

internal/controller/appconfigurationprovider_controller.go
RichardChen820: Updated

internal/controller/appconfigurationprovider_controller.go
RichardChen820: `existingSecrets` is  calculated eveytime of reconciliation

internal/controller/appconfigurationprovider_controller.go
juniwang: thanks

internal/controller/appconfigurationprovider_controller.go
juniwang: Got it, thanks

juniwang (approved):

PR #39, opened 2024-04-30
Workload identity support namespaced service account

RichardChen820:
Workload identity support namespaced service account


internal/loader/configuration_client_manager.go
```diff
@@ -461,3 +474,59 @@ func Jitter(duration time.Duration) time.Duration {
...
+	err = client.Get(ctx, types.NamespacedName{Namespace: serviceAccountNamespace, Name: serviceAccountName}, serviceAccountObj)
+	if err != nil {
+		return nil, err
+	}
+
+	clientId := serviceAccountObj.ObjectMeta.Annotations[AnnotationClientID]
+	tenantId := serviceAccountObj.Annotations[AnnotationTenantID]
+	getAssertionFunc := newGetAssertionFunc(serviceAccountNamespace, serviceAccountName)
+
+	clientAssertionCredential, err := azidentity.NewClientAssertionCredential(tenantId, clientId, getAssertionFunc, nil)
```
linglingye001, line 499: I am not sure the usage of `clientAssertionCredential`,for authenticating both service principal and workload identity?

internal/loader/configuration_client_manager.go
```diff
@@ -461,3 +474,59 @@ func Jitter(duration time.Duration) time.Duration {
...
+		if err != nil {
+			return "", err
+		}
+
+		kubeClient, err := kubernetes.NewForConfig(cfg)
+		if err != nil {
+			return "", err
+		}
+
+		token, err := kubeClient.CoreV1().ServiceAccounts(serviceAccountNamespace).CreateToken(ctx, serviceAccountName, &authv1.TokenRequest{
```
linglingye001, line 521: Will the service account token expire?

internal/loader/configuration_client_manager.go
```diff
@@ -461,3 +474,59 @@ func Jitter(duration time.Duration) time.Duration {
...
+		return nil, err
+	}
+
+	serviceAccountObj := &corev1.ServiceAccount{}
+	err = client.Get(ctx, types.NamespacedName{Namespace: serviceAccountNamespace, Name: serviceAccountName}, serviceAccountObj)
+	if err != nil {
+		return nil, err
+	}
+
+	clientId := serviceAccountObj.ObjectMeta.Annotations[AnnotationClientID]
```
linglingye001, line 495: Can we add the check for `AnnotationClientID` and `AnnotationTenantID`? They may not exist.

internal/loader/configuration_client_manager.go
```diff
@@ -492,11 +493,22 @@ func newClientAssertionCredential(ctx context.Context, serviceAccountName string
...
+	if !ok {
+		return nil, fmt.Errorf("no tenant ID specified. Check pod configuration or set TenantID in the options")
+	}
+
+	if _, ok := serviceAccountObj.Annotations[AnnotationClientID]; !ok {
+		return nil, fmt.Errorf("service account %s/%s does not have annotation %s", serviceAccountNamespace, serviceAccountName, AnnotationClientID)
+	}
+
+	if _, ok := serviceAccountObj.Annotations[AnnotationTenantID]; ok {
+		tenantId = serviceAccountObj.Annotations[AnnotationTenantID]
```
linglingye001, line 506: What's the difference bewteen the `AzureTenantId` from pod config and `AnnotationTenantID` from service account?

internal/loader/configuration_client_manager.go
```diff
@@ -492,11 +493,22 @@ func newClientAssertionCredential(ctx context.Context, serviceAccountName string
 		return nil, err
 	}
 
-	clientId := serviceAccountObj.ObjectMeta.Annotations[AnnotationClientID]
-	tenantId := serviceAccountObj.Annotations[AnnotationTenantID]
+	tenantId, ok := os.LookupEnv(strings.ToUpper(AzureTenantId))
```
linglingye001, line 496: Why the `tenantId` is required?

deploy/templates/deployment.yaml
```diff
@@ -51,6 +51,10 @@ spec:
         - name: AZURE_TENANT_ID
           value: {{ .Values.env.azureTenantId }}
         {{- end }}
+        - name: WORKLOAD_IDENTITY_ENABLED
+          value: {{ .Values.workloadIdentity.enabled }}
+        - name: WORKLOAD_IDENTITY_CUSTOM_SERVICE_ACCOUNT_ONLY
+          value: {{ .Values.workloadIdentity.allowCustomServiceAccount }}
```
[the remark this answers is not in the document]
RichardChen820: Updated

internal/loader/configuration_client_manager.go
RichardChen820: Good question, the `tenantId` is there if enable the cluster workload identity, but need confirm it.

internal/loader/configuration_client_manager.go
RichardChen820: It could expire, but since we reuse the appconfig.client, once the tokenCredential expires, but will try to acquire new AAD token, then it will call the callback to get a new service account token as well, so we do not need to worry about service account expires or not.

internal/loader/configuration_client_manager.go
RichardChen820: You can imagine that the `AzureTenantId` from pod is global, the `AnnotationTenantID` is for each specific service account. If users uses the same tenantID for all service account, it just need to use the global one, no need to specify in every service account

internal/loader/configuration_client_manager.go
RichardChen820: Confirmed, once the workload identity is enabled, the `tenantId` should be there, and the existing way we support for workload identity also reply on this `tenantId `

linglingye001: If namespaced service account is used, is the value of `tenantId` same as `AnnotationTenantID` from service account? If customers need to explicitly set this value, will users from other namespace can get the `tenantId` through `kubectl describe po k8sproviderpodxxxx`？

RichardChen820: I don't quite get your question. `tenantId` if for global, each service account could overwrite is by setting `AnnotationTenantID`

linglingye001 (approved):

internal/controller/utils.go
```diff
@@ -237,12 +240,36 @@ func verifyRefreshInterval(interval string, allowedMinimalRefreshInterval time.D
...
+	if !strings.EqualFold(os.Getenv(WorkloadIdentityEnabled), "true") {
+		return loader.NewArgumentError("auth.workloadIdentity", fmt.Errorf("workloadIdentity is not enabled"))
 	}
 
-	if workloadIdentity.ManagedIdentityClientId != nil && workloadIdentity.ManagedIdentityClientIdReference != nil {
-		return loader.NewArgumentError("auth.workloadIdentity", fmt.Errorf("only one of managedIdentityClientId and managedIdentityClientIdReference is allowed"))
+	var authCount int = 0
+
+	if workloadIdentity.ManagedIdentityClientId != nil {
+		if strings.EqualFold(os.Getenv(WorkloadIdentityDisableGlobalServiceAccount), "true") {
```
juniwang, line 250: could you remind me again what's the issue to allow `ManageIdentityClientId` when global service account is disabled?

internal/controller/utils.go
```diff
@@ -237,12 +240,36 @@ func verifyRefreshInterval(interval string, allowedMinimalRefreshInterval time.D
...
+			return loader.NewArgumentError("auth.workloadIdentity.managedIdentityClientId", fmt.Errorf("managedIdentityClientId is not allowed since only custom service account is allowed"))
```
juniwang, line 251: should we avoid "custom service account"? some like  "since global service account is disabled" to keep consistent with the installation flag.

internal/loader/configuration_client_manager.go
juniwang: to be honest, I cannot think of a case when a different tenant id is used. Technically it has to be the same tenantId.

internal/loader/configuration_client_manager.go
```diff
@@ -461,3 +475,70 @@ func Jitter(duration time.Duration) time.Duration {
...
+	clientAssertionCredential, err := azidentity.NewClientAssertionCredential(tenantId, serviceAccountObj.Annotations[AnnotationClientID], getAssertionFunc, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	return clientAssertionCredential, nil
+}
+
+func newGetAssertionFunc(serviceAccountNamespace string, serviceAccountName string) func(ctx context.Context) (string, error) {
+	audiences := []string{AzureDefaultAudience}
```
juniwang, line 520: the name is a little misleading. it's not default audience of Azure. I believe https://management.azure.com is the actual default for control plane. `ApiTokenExchangeAudience`?

internal/controller/utils.go
RichardChen820: The way of using `ManageIdentityClientId` is tied to the global service account, customer has the security concern that it cross the namespace security boundary, namespace owners can easily use the `ManageIdentityClientId` from other team(namespace) to access the App Config which they should not have the permission to access. 

This PR is aim to add the namespaced service account support to eliminate the security concern of customer. At the meanwhile, we give an option to disable the global service account to prevent them from using the risky way.

internal/loader/configuration_client_manager.go
RichardChen820: Make sense, updated.

internal/controller/utils.go
RichardChen820: Updated

internal/loader/configuration_client_manager.go
RichardChen1122: I do a bit change to this, hope this make sense to you

internal/controller/utils.go
juniwang: I'm actually asking why the `ManageIdentityClientId` is tied to the global service account, instead of allowing it to be used by customer specified service account too.

deploy/templates/clusterrole.yaml
```diff
@@ -53,6 +53,20 @@ rules:
...
+- apiGroups:
+  - ""
+  resources:
+  - serviceaccounts
+  verbs:
+  - get
+- apiGroups:
+  - ""
+  resources:
+  - serviceaccounts/token
```
juniwang, line 66: are these new permissions required only when `disableGlobalServiceAccount` is true?

RichardChen820: No, when `disableGlobalServiceAccount ` is false, customers can still use namespaced service account, as long as they are allowed to passing their own service account, we need these role.

internal/controller/utils.go
RichardChen820: Are you saying why not support something like this?

```
auth:
   workloadIdentity:
      serviceAccountName: xxx
     ManageIdentityClientId: xxx
```

deploy/templates/clusterrole.yaml
juniwang: make sense

internal/controller/utils.go
juniwang: yes. what's the reason behind?

RichardChen820: It will be ambiguous for users which one is exactly taking effect. If we allow this, there are some implicit assumptions involved. If both `serviceAccountName` and `managedIdentityClientId` are set, then managedIdentityClientId is assumed to be bound to the specified serviceAccountName. However, if serviceAccountName is not set, it may give the impression that serviceAccountName has a default value, which is the global service account. But this is not the case; namespaced service accounts and global service accounts are not the same and should not be conflated. This could easily lead to customer confusion.

deploy/templates/serviceaccount.yaml
```diff
@@ -9,7 +9,7 @@ metadata:
     {{- if .Values.serviceAccount.annotations }}
     {{ toYaml .Values.serviceAccount.annotations . | nindent 4 }}
     {{- end }}
-    {{- if eq .Values.workloadIdentity.enabled true }}
+    {{- if and (.Values.workloadIdentity.enabled) (not .Values.workloadIdentity.disableGlobalServiceAccount) }}
```
juniwang, line 12: will this will this disable the service account? Looks like it only stops annotating with client id

RichardChen820: We can't actually get rid of this service account since it bind with clusterRoles, we still need it for accessing resources. Remove this annotation can make the service account lose the ability to use the workload identity since the Service Account Token Volume Projection can't work won't work without this annotation

juniwang: got it ,thanks

juniwang (approved):

PR #45, opened 2024-05-31
Deduplicate feature flags

linglingye001:


version.json
```diff
@@ -1,3 +1,3 @@
 {
-    "version":"1.3.1"
```
RichardChen820, line 2: Don't update the version in it.

RichardChen820: And I think this fix can be rollout in 2.0.0 as well

PR #46, opened 2024-06-04
Add node affinity and toleration configuration

RichardChen820:
Add node affinity and toleration configuration

### DIFF
--- a/deploy/templates/deployment.yaml
+++ b/deploy/templates/deployment.yaml
@@ -24,13 +24,37 @@ spec:
       nodeSelector:
         {{- toYaml .Values.nodeSelector | nindent 8 }}
       {{- end }}
-      {{- if .Values.affinity }}
       affinity:
+      {{- if .Values.affinity }}
         {{- toYaml .Values.affinity | nindent 8 }}
+      {{- else }}
+        nodeAffinity:
+          preferredDuringSchedulingIgnoredDuringExecution:
+          - weight: 100
+            preference:
+              matchExpressions:
+              - key: kubernetes.azure.com/mode
+                operator: In
+                values:
+                - system
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+            - matchExpressions:
+              - key: kubernetes.io/os
+                operator: NotIn
+                values:
+                - windows
+              - key: type
+                operator: NotIn
+                values:
+                - virtual-kubelet
       {{- end }}
-      {{- if .Values.tolerations }}
       tolerations:
-        {{- toYaml .Values.tolerations | nindent 8 }}
+      {{- if .Values.tolerations }}
+        {{- toYaml .Values.tolerations | nindent 6 }}
+      {{- else }}
+      - key: "CriticalAddonsOnly"
+        operator: "Exists"
       {{- end }}
       containers:
       - args:
--- a/deploy/parameter/helm-values.yaml
+++ b/deploy/parameter/helm-values.yaml
@@ -60,15 +60,7 @@ env:
 
 # Pod scheduling preferences
 # ref: https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#affinity-and-anti-affinity
-affinity:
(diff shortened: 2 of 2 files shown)


deploy/templates/deployment.yaml
```diff
@@ -24,13 +24,37 @@ spec:
       nodeSelector:
         {{- toYaml .Values.nodeSelector | nindent 8 }}
       {{- end }}
-      {{- if .Values.affinity }}
       affinity:
+      {{- if .Values.affinity }}
         {{- toYaml .Values.affinity | nindent 8 }}
+      {{- else }}
```
linglingye001, line 30: What's the default `nodeAffinity` used for?

deploy/templates/deployment.yaml
```diff
@@ -24,13 +24,37 @@ spec:
...
+                values:
+                - virtual-kubelet
       {{- end }}
-      {{- if .Values.tolerations }}
       tolerations:
-        {{- toYaml .Values.tolerations | nindent 8 }}
+      {{- if .Values.tolerations }}
+        {{- toYaml .Values.tolerations | nindent 6 }}
+      {{- else }}
+      - key: "CriticalAddonsOnly"
```
linglingye001, line 56: Why need this toleration

deploy/templates/deployment.yaml
```diff
@@ -24,13 +24,37 @@ spec:
...
         {{- toYaml .Values.nodeSelector | nindent 8 }}
       {{- end }}
-      {{- if .Values.affinity }}
       affinity:
+      {{- if .Values.affinity }}
         {{- toYaml .Values.affinity | nindent 8 }}
+      {{- else }}
+        nodeAffinity:
+          preferredDuringSchedulingIgnoredDuringExecution:
+          - weight: 100
```
juniwang, line 33: `weight` seems to be unnecessary if only one preference

deploy/templates/deployment.yaml
```diff
@@ -24,13 +24,37 @@ spec:
...
+            nodeSelectorTerms:
+            - matchExpressions:
+              - key: kubernetes.io/os
+                operator: NotIn
+                values:
+                - windows
+              - key: type
+                operator: NotIn
+                values:
+                - virtual-kubelet
```
juniwang, line 50: why restricts `virtual-kubelet`? any issue found/reported?

deploy/templates/deployment.yaml
```diff
@@ -24,13 +24,37 @@ spec:
...
       {{- end }}
-      {{- if .Values.tolerations }}
       tolerations:
-        {{- toYaml .Values.tolerations | nindent 8 }}
+      {{- if .Values.tolerations }}
+        {{- toYaml .Values.tolerations | nindent 6 }}
```
juniwang, line 54: the original indentation `8` seems to be right.

RichardChen820: It's an array. The format convention should be:

```
tolerations:
- xx
- xx
```

not 

```
tolerations:
  - xx
  - xx
```

deploy/templates/deployment.yaml
RichardChen820: It's recommended by the AKS, and it's make sense to us, since there's no fair reason to run k8s provider on a `virtual-kubelet` node

deploy/templates/deployment.yaml
RichardChen820: Prefer running the provider in system node, not running it in windows node and virtual kublet node.

deploy/templates/deployment.yaml
RichardChen820: AKS system node usually has "CriticalAddonsOnly", tolerate it that the provider can run on a system node.

deploy/templates/deployment.yaml
RichardChen820: This is a required field, range 1-100

linglingye001 (approved):

juniwang (approved):

PR #47, opened 2024-06-05
Add Correlation Context header for extension

RichardChen820:
Add Correlation Context header for extension

### DIFF
--- a/deploy/parameter/helm-values.yaml
+++ b/deploy/parameter/helm-values.yaml
@@ -15,6 +15,9 @@ imagePullSecrets: []
 nameOverride: "appconfig-provider"
 fullnameOverride: "az-appconfig-k8s-provider"
 
+requestTracing:
+  enabled: true
+
 workloadIdentity:
   enabled: true
   disableGlobalServiceAccount: false
--- a/deploy/templates/deployment.yaml
+++ b/deploy/templates/deployment.yaml
@@ -55,6 +55,13 @@ spec:
           value: "{{ .Values.workloadIdentity.enabled }}"
         - name: WORKLOAD_IDENTITY_DISABLE_GLOBAL_SERVICE_ACCOUNT
           value: "{{ .Values.workloadIdentity.disableGlobalServiceAccount }}"
+        {{- if and (hasKey .Values "global") 
+            (hasKey .Values.global "azure")
+            (hasKey .Values.global.azure "extension")
+            (hasKey .Values.global.azure.extension "name") }}
+        - name: AZURE_EXTENSION_CONTEXT
+          value: UseExtension
+        {{- end }}
         livenessProbe:
           httpGet:
             path: /healthz
--- a/deploy/templates/deployment.yaml
+++ b/deploy/templates/deployment.yaml
@@ -62,6 +62,8 @@ spec:
         - name: AZURE_EXTENSION_CONTEXT
           value: UseExtension
         {{- end }}
+        - name: REQUEST_TRACING_ENABLED
+          value: "{{ .Values.requestTracing.enabled }}"
         livenessProbe:
           httpGet:
             path: /healthz
--- a/internal/controller/appconfigurationprovider_controller.go
+++ b/internal/controller/appconfigurationprovider_controller.go
@@ -224,6 +224,10 @@ func (reconciler *AzureAppConfigurationProviderReconciler) Reconcile(ctx context
 		retriever = reconciler.Retriever
 	}
 
+	ctx = context.WithValue(ctx, loader.RequestTracingKey, loader.RequestTracing{
+		IsStartUp: reconciler.ProvidersReconcileState[req.NamespacedName].ConfigMapResourceVersion == nil,
+	})
+
 	// Initialize the processor setting in this reconcile
 	processor := &AppConfigurationProviderProcessor{
 		Context:                 ctx,
--- a/internal/controller/appconfigurationprovider_controller.go
+++ b/internal/controller/appconfigurationprovider_controller.go
@@ -126,7 +126,7 @@ func (reconciler *AzureAppConfigurationProviderReconciler) Reconcile(ctx context
 
 	err = verifyObject(provider.Spec)
 	if err != nil {
-		reconciler.logAndSetFailStatus(ctx, err, provider)
+		reconciler.logAndSetFailStatus(provider, err)
 		return reconcile.Result{Requeue: false}, nil
 	}
 
@@ -134,7 +134,7 @@ func (reconciler *AzureAppConfigurationProviderReconciler) Reconcile(ctx context
 	isExisting := false
 	_, err = reconciler.verifyTargetObjectExistence(ctx, provider, &existingConfigMap)
 	if err != nil {
-		reconciler.logAndSetFailStatus(ctx, err, provider)
+		reconciler.logAndSetFailStatus(provider, err)
 		return reconcile.Result{Requeue: true, RequeueAfter: RequeueReconcileAfter}, nil
 	}
 
@@ -148,7 +148,7 @@ func (reconciler *AzureAppConfigurationProviderReconciler) Reconcile(ctx context
 		}
 		isExisting, err = reconciler.verifyTargetObjectExistence(ctx, provider, &existingSecret)
 		if err != nil {
-			reconciler.logAndSetFailStatus(ctx, err, provider)
+			reconciler.logAndSetFailStatus(provider, err)
 			return reconcile.Result{Requeue: true, RequeueAfter: RequeueReconcileAfter}, nil
 		}
 		if isExisting {
@@ -166,7 +166,7 @@ func (reconciler *AzureAppConfigurationProviderReconciler) Reconcile(ctx context
 				}
 				isExisting, err = reconciler.verifyTargetObjectExistence(ctx, provider, &existingSecret)
 				if err != nil {
-					reconciler.logAndSetFailStatus(ctx, err, provider)
+					reconciler.logAndSetFailStatus(provider, err)
 					return reconcile.Result{Requeue: true, RequeueAfter: RequeueReconcileAfter}, nil
 				}
 				if isExisting {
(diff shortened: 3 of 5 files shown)


internal/controller/appconfigurationprovider_controller.go
```diff
@@ -311,7 +311,6 @@ func (reconciler *AzureAppConfigurationProviderReconciler) verifyTargetObjectExi
 }
 
 func (reconciler *AzureAppConfigurationProviderReconciler) logAndSetFailStatus(
-	ctx context.Context,
```
RichardChen820, line 314: Just remove the unused parameter here.

internal/loader/configuration_setting_loader.go
```diff
@@ -743,3 +758,40 @@ func MergeSecret(secret map[string]corev1.Secret, newSecret map[string]corev1.Se
 
 	return nil
 }
+
+func createCorrelationContextHeader(provider acpv1.AzureAppConfigurationProvider) http.Header {
+	header := http.Header{}
+	output := make([]string, 0)
+
+	if provider.Spec.Configuration.Refresh != nil {
```
linglingye001, line 766: `Spec.Configuration` can be nil

linglingye001: `Refresh.Enabled`  can be false, need to check it when `Configuration.Refresh` section has been set.

internal/loader/configuration_setting_loader.go
```diff
@@ -455,6 +461,15 @@ func (csl *ConfigurationSettingLoader) ExecuteFailoverPolicy(ctx context.Context
 		return nil, fmt.Errorf("no client is available to connect to the target App Configuration store")
 	}
 
+	if value, ok := os.LookupEnv(RequestTracingEnabled); ok {
+		enabled := true
+		enabled, _ = strconv.ParseBool(value)
+
+		if enabled {
```
linglingye001, line 468: It can be simplified to ` if enabled, _ := strconv.ParseBool(value); enabled {`

deploy/templates/deployment.yaml
```diff
@@ -55,6 +55,15 @@ spec:
           value: "{{ .Values.workloadIdentity.enabled }}"
         - name: WORKLOAD_IDENTITY_DISABLE_GLOBAL_SERVICE_ACCOUNT
           value: "{{ .Values.workloadIdentity.disableGlobalServiceAccount }}"
+        {{- if and (hasKey .Values "global") 
+            (hasKey .Values.global "azure")
```
linglingye001, line 59: Not sure, will these values like `global`, `azure`, `extension` and `name` be added during the extension installation?

internal/loader/configuration_setting_loader.go
```diff
@@ -743,3 +759,51 @@ func MergeSecret(secret map[string]corev1.Secret, newSecret map[string]corev1.Se
...
+			output = append(output, "RefreshesKeyVault")
+		}
+	}
+
+	if provider.Spec.FeatureFlag != nil {
+		output = append(output, "UsesFeatureFlag")
+
+		if provider.Spec.FeatureFlag.Refresh != nil &&
+			provider.Spec.FeatureFlag.Refresh.Enabled {
+			output = append(output, "RefreshesFeatureFlag")
```
zhenlan, line 785: Feature flag by nature should be refreshed. This is not useful to collect.

The purpose of `RefreshesKeyVault` is to understand how many stores use appconfig for automatic secret rotation.

internal/loader/configuration_setting_loader.go
```diff
@@ -743,3 +759,51 @@ func MergeSecret(secret map[string]corev1.Secret, newSecret map[string]corev1.Se
 
 	return nil
 }
+
+func createCorrelationContextHeader(provider acpv1.AzureAppConfigurationProvider, clientManager ClientManager) http.Header {
```
zhenlan, line 763: Can we also report `RequestType`, `Host`, and `Filter`?

https://github.com/Azure/AppConfiguration-DotnetProvider/blob/main/src/Microsoft.Extensions.Configuration.AzureAppConfiguration/Constants/RequestTracingConstants.cs#L23

deploy/templates/deployment.yaml
RichardChen820: Yes， this is what will be passed by the extension

linglingye001:
Can we update `README` for Data collection? https://github.com/Azure/AppConfiguration-DotnetProvider?tab=readme-ov-file#data-collection

internal/loader/configuration_setting_loader.go
RichardChen820: @zhenlan  Added the `RequestType` and `Host`. In terms of the `Filter`, it's not cheap to report it from k8s provider, not trivial in K8s provider, two main reasons:

1. We currently don't parse the feature flags in K8s provider, we just get them and put into configMap, if we need to report the filters, we need to parse the feature flag values to know what filters are set. 
2. K8s provider manages all yaml resouces, we need to let the controller can remember/manage all filter usage information for each resource.

How do you think about it? We add the `Filter` for now, or add in the future on demand?

internal/loader/request_tracing.go
```diff
@@ -0,0 +1,78 @@
...
+		if tracing.(RequestTracing).IsStartUp {
+			output = append(output, "RequestType=StartUp")
+		} else {
+			output = append(output, "RequestType=Watch")
+		}
+	}
+
+	if provider.Spec.Configuration.Refresh != nil &&
+		provider.Spec.Configuration.Refresh.Enabled {
+		output = append(output, "RefreshesKeyValue")
```
zhenlan, line 42: Not needed. We don't have this in .NET either.

internal/loader/request_tracing.go
```diff
@@ -0,0 +1,78 @@
...
+		output = append(output, "UsesKeyVault")
+
+		if provider.Spec.Secret.Refresh != nil &&
+			provider.Spec.Secret.Refresh.Enabled {
+			output = append(output, "RefreshesKeyVault")
+		}
+	}
+
+	if provider.Spec.FeatureFlag != nil {
+		output = append(output, "UsesFeatureFlag")
```
zhenlan, line 55: Not needed. We can tell from requests.

internal/loader/configuration_setting_loader.go
zhenlan: If we are going to do dedup, we may parse it already. cc: @linglingye001

RichardChen820:
> Can we update `README` for Data collection? https://github.com/Azure/AppConfiguration-DotnetProvider?tab=readme-ov-file#data-collection

Yes, I will update it another PR

internal/loader/configuration_setting_loader.go
linglingye001: We don't parse feature flag now, the implementation of feature flag deduplication is based on feature flag key.

RichardChen820: I think Zhenlan's view is that we need to dedup by id/name, then drives us having to parse it.

internal/controller/appconfigurationprovider_controller.go
```diff
@@ -224,6 +224,10 @@ func (reconciler *AzureAppConfigurationProviderReconciler) Reconcile(ctx context
 		retriever = reconciler.Retriever
 	}
 
+	ctx = context.WithValue(ctx, loader.RequestTracingKey, loader.RequestTracing{
+		IsStartUp: reconciler.ProvidersReconcileState[req.NamespacedName].ConfigMapResourceVersion != nil,
```
juniwang, line 228: Is ConfigMapResourceVersion `nil` when starting up?

internal/loader/request_tracing.go
juniwang: I'm wondering why there is `RefreshesKeyVault` in dotnet provider but not `RefreshesKeyValue`?

zhenlan: We know customers like dynamic config. We don't need to collect telemetry to know it.

I explained the reason for key vault refresh here https://github.com/Azure/AppConfiguration-KubernetesProvider/pull/47#discussion_r1645291492.

internal/controller/appconfigurationprovider_controller.go
RichardChen820: My bad.. updated

internal/loader/configuration_setting_loader.go
```diff
@@ -88,6 +91,7 @@ const (
 	CertTypePfx                           string = "application/x-pkcs12"
 	TlsKey                                string = "tls.key"
 	TlsCrt                                string = "tls.crt"
+	RequestTracingEnabled                 string = "REQUEST_TRACING_ENABLED"
```
zhenlan, line 94: How does a customer disable the tracing? What about the AKS extension customer?

RichardChen820: They can set `requestTracing.enabled=false` while installling the proivder. Same to the extension customer.

zhenlan: You mean in the yaml file?

deploy/parameter/helm-values.yaml
```diff
@@ -15,6 +15,9 @@ imagePullSecrets: []
 nameOverride: "appconfig-provider"
 fullnameOverride: "az-appconfig-k8s-provider"
 
+requestTracing:
```
RichardChen820, line 18: Set requestTracing.enabled at installation time

internal/loader/configuration_setting_loader.go
RichardChen820: @zhenlan I mean here https://github.com/Azure/AppConfiguration-KubernetesProvider/pull/47/files#r1652128093

zhenlan: This is the parameter a user pass to the helm install? If so, how can an extension user pass the parameter?

RichardChen820: Extension essentially use helm to do the installation, it supports all the helm values. You can consider there are no difference

zhenlan: I understand. My question is how an extension user can customize the helm install command and pass the extra parameter.

RichardChen820: LIke how replicaCount can be set:
https://learn.microsoft.com/en-us/azure/aks/azure-app-configuration-settings#configure-the-replica-count

User can customize use this command while install the extension.

```

az k8s-extension create --cluster-type managedClusters --cluster-name myAKSCluster --resource-group myResourceGroup --name appconfigurationkubernetesprovider --extension-type Microsoft.AppConfiguration --auto-upgrade-minor-version true --configuration-settings "requestTracing.enabled=false"
```

zhenlan: Okay, so it's a different command. I am curious how you plan to explain this to customers.

juniwang (approved):

PR #53, opened 2024-07-05
Refresh secret when data change

linglingye001:


internal/controller/appconfigurationprovider_controller.go
```diff
@@ -426,35 +426,38 @@ func (reconciler *AzureAppConfigurationProviderReconciler) createOrUpdateSecrets
...
+				reconciler.logAndSetFailStatus(provider, err)
+				return reconcile.Result{Requeue: true, RequeueAfter: RequeueReconcileAfter}, err
+			}
 
-		reconciler.ProvidersReconcileState[namespacedName].ExistingSecretReferences[secretObj.Name].SecretResourceVersion = secretObj.ResourceVersion
-		klog.V(5).Infof("Secret %q in %q namespace is %s", secretObj.Name, secretObj.Namespace, string(operationResult))
+			reconciler.ProvidersReconcileState[namespacedName].ExistingSecretReferences[secretObj.Name].SecretResourceVersion = secretObj.ResourceVersion
+			reconciler.ProvidersReconcileState[namespacedName].ExistingSecretReferences[secretObj.Name].UpdateNeeded = false
+			klog.V(5).Infof("Secret %q in %q namespace is %s", secretObj.Name, secretObj.Namespace, string(operationResult))
+		}
```
RichardChen820, line 460: Add a v(5) log here if the secret is not updated.

internal/controller/appconfigurationprovider_controller_test.go
```diff
@@ -608,126 +608,6 @@ var _ = Describe("AppConfiguationProvider controller", func() {
 
 			_ = k8sClient.Delete(ctx, configProvider)
 		})
-
-		It("Should refresh secret", func() {
```
RichardChen820, line 612: What is the reason of removing this test

internal/loader/keyvault_reference_resolver.go
```diff
@@ -28,6 +28,7 @@ type KeyVaultSecretUriSegment struct {
 	HostName      string
```
RichardChen820, line 28: This is longer just indicates the UriSegment of the Secret, how about rename this type to `KeyVaultSecretMetadata` ?

internal/controller/processor.go
```diff
@@ -201,6 +210,7 @@ func (processor *AppConfigurationProviderProcessor) processSecretReferenceRefres
 
 	// Only resolve the secret references that not specified the secret version
 	secretReferencesToSolve := make(map[string]*loader.TargetSecretReference)
+	cachedSecretReferences := make(map[string]*loader.TargetSecretReference)
```
RichardChen820, line 213: This variable seems useless, do we really need it?

internal/controller/processor.go
```diff
@@ -332,3 +354,28 @@ func (processor *AppConfigurationProviderProcessor) calculateRequeueAfterInterva
 
 	return requeueAfterInterval
 }
+
+func checkSecretVersion(existingSecretReferences map[string]*loader.TargetSecretReference, latestSecretReferences map[string]*loader.TargetSecretReference) {
```
RichardChen820, line 358: This method is not well self-explained, check secret version for what?  I would suggest this method has a boolean type return value, and then update the value to `secretReference.UpdateNeeded`

internal/controller/processor.go
```diff
@@ -70,6 +70,9 @@ func (processor *AppConfigurationProviderProcessor) processFullReconciliation()
 		return err
 	}
 	processor.Settings = updatedSettings
+	for _, reference := range updatedSettings.SecretReferences {
+		reference.UpdateNeeded = true
+	}
 	processor.ReconciliationState.ExistingSecretReferences = updatedSettings.SecretReferences
```
RichardChen820, line 76: Since the SecretReferences now involves the information of secret now, we should set it to ExistingSecretReferences in the `Finish()` step now, otherwise it might mismatch with the actual version in the K8s secret.

internal/controller/processor.go
```diff
@@ -160,6 +163,12 @@ func (processor *AppConfigurationProviderProcessor) processKeyValueRefresh(exist
 	}
 
 	processor.Settings = keyValueRefreshedSettings
+	checkSecretVersion(processor.ReconciliationState.ExistingSecretReferences, keyValueRefreshedSettings.SecretReferences)
+	for secretName, reference := range keyValueRefreshedSettings.SecretReferences {
+		if _, ok := processor.ReconciliationState.ExistingSecretReferences[secretName]; ok {
+			reference.SecretResourceVersion = processor.ReconciliationState.ExistingSecretReferences[secretName].SecretResourceVersion
+		}
+	}
 	reconcileState.ExistingSecretReferences = keyValueRefreshedSettings.SecretReferences
```
RichardChen820, line 172: Same with what I comment on the line 73

internal/controller/processor.go
```diff
@@ -70,6 +70,9 @@ func (processor *AppConfigurationProviderProcessor) processFullReconciliation()
 		return err
 	}
 	processor.Settings = updatedSettings
+	for _, reference := range updatedSettings.SecretReferences {
```
RichardChen820, line 73: Current all the version check are in the processor, result in we need to set the `updateNeeded` everywhere for the secret references, could we put the version check in the controller right before trying to update the Secret? I think that would be more clear.

internal/controller/processor.go
linglingye001: Currently  `ResolveSecretReferences` doesn't pass/return new parameters for `secretID` of latest secret, it just update the `secretID` in `kvReferencesToResolve` . So need to cache secretRef before resolving.

internal/controller/processor.go
```diff
@@ -201,17 +199,26 @@ func (processor *AppConfigurationProviderProcessor) processSecretReferenceRefres
 
 	// Only resolve the secret references that not specified the secret version
 	secretReferencesToSolve := make(map[string]*loader.TargetSecretReference)
+	cachedSecretReferences := make(map[string]*loader.TargetSecretReference)
```
RichardChen820, line 202: This name is really confusion-leading, what you do here is to deep copy the `reconcileState.ExistingSecretReferences`, right?

internal/loader/configuration_setting_loader.go
```diff
@@ -50,7 +50,7 @@ type TargetKeyValueSettings struct {
 
 type TargetSecretReference struct {
 	Type                  corev1.SecretType
-	UriSegments           map[string]KeyVaultSecretUriSegment
+	UriSegments           map[string]KeyVaultSecretMetadata
```
RichardChen820, line 53: The name `UriSegments` does not make sense now, how about updated to `SecretMetadatas`

internal/controller/processor.go
```diff
@@ -201,17 +199,26 @@ func (processor *AppConfigurationProviderProcessor) processSecretReferenceRefres
...
 				if secretReferencesToSolve[secretName] == nil {
 					secretReferencesToSolve[secretName] = &loader.TargetSecretReference{
 						Type:        reference.Type,
-						UriSegments: make(map[string]loader.KeyVaultSecretUriSegment),
+						UriSegments: make(map[string]loader.KeyVaultSecretMetadata),
 					}
 				}
 				secretReferencesToSolve[secretName].UriSegments[key] = uriSegment
 			}
+
```
RichardChen820, line 214: I believe you are trying to do map deepcopy here right?

internal/controller/processor.go
```diff
@@ -270,6 +283,9 @@ func (processor *AppConfigurationProviderProcessor) shouldReconcile(
 
 func (processor *AppConfigurationProviderProcessor) Finish() (ctrl.Result, error) {
 	processor.ReconciliationState.Generation = processor.Provider.Generation
+	if processor.RefreshOptions.SecretSettingPopulated {
```
RichardChen820, line 286: Is there any sepcial reason of not putting this block into the line295?

internal/controller/utils.go
```diff
@@ -302,3 +302,41 @@ func verifySelectorObject(selector acpv1.Selector) error {
 
 	return nil
 }
+
+func checkAndUpdateSecretRef(existingSecretReferences map[string]*loader.TargetSecretReference, latestSecretReferences map[string]*loader.TargetSecretReference, shouldReconcile bool) map[string]bool {
```
RichardChen820, line 306: Could we let the method check `secretUpdateNeeded` one by one?

internal/controller/appconfigurationprovider_controller.go
```diff
@@ -269,15 +269,15 @@ func (reconciler *AzureAppConfigurationProviderReconciler) Reconcile(ctx context
...
+		result, err := reconciler.createOrUpdateSecrets(ctx, provider, processor)
 		if err != nil {
 			return result, nil
 		}
 	}
 
 	// Expel the secrets which are no longer selected by the provider.
 	if provider.Spec.Secret == nil || processor.RefreshOptions.SecretSettingPopulated {
-		result, err := reconciler.expelRemovedSecrets(ctx, provider, existingSecrets, processor.ReconciliationState.ExistingSecretReferences)
+		result, err := reconciler.expelRemovedSecrets(ctx, provider, existingSecrets, processor.Settings.SecretReferences)
```
RichardChen820, line 280: I didn't get the reason of doing this change.

internal/controller/appconfigurationprovider_controller.go
```diff
@@ -425,36 +425,41 @@ func (reconciler *AzureAppConfigurationProviderReconciler) createOrUpdateSecrets
...
-	for secretName, secret := range settings.SecretSettings {
-		secretObj := &corev1.Secret{
-			ObjectMeta: metav1.ObjectMeta{
-				Name:      secretName,
-				Namespace: provider.Namespace,
-			},
-			Type: secret.Type,
-		}
+	secretToUpdate := checkAndUpdateSecretRef(reconciler.ProvidersReconcileState[namespacedName].ExistingSecretReferences, processor.Settings.SecretReferences, processor.ShouldReconcile)
+	for secretName, secret := range processor.Settings.SecretSettings {
```
RichardChen820, line 429: How about doing this:

```
for secretName, secret := range processor.Settings.SecretSettings {
   if (!shouldCreateOrUpdate()){
     continue;
   }

    // do create or update
     .....
}

```

internal/controller/processor.go
linglingye001: Not all reconcile triggered by refresh, when `shouldReconcile=true` such as delete configmap/secret, we need to maintain `existingSecretReferences` when `secret` section specified.

internal/controller/processor.go
linglingye001: yes, how about `copiedSecretReferences`?

internal/controller/processor.go
linglingye001: yes, do you mean we need to fave a new `loop` for map deepcopy?

internal/controller/appconfigurationprovider_controller.go
```diff
@@ -269,15 +269,15 @@ func (reconciler *AzureAppConfigurationProviderReconciler) Reconcile(ctx context
 			}
```
RichardChen820, line 269: Here is the right place to check `shouldCreateOrUpdate`

```
for name := range processor.Settings.SecretSettings {
     // check the target object existence

    if (shouldCreateOrUpdate){
               result, err := reconciler.createOrUpdateSecret(ctx, provider, SecretSetting)
		if err != nil {
			return result, nil
		}
    }

}
```

RichardChen820: We'd better not let one method to perform all secrets update. We check and update for each.

internal/controller/appconfigurationprovider_controller.go
```diff
@@ -425,36 +425,43 @@ func (reconciler *AzureAppConfigurationProviderReconciler) createOrUpdateSecrets
...
 
-		reconciler.ProvidersReconcileState[namespacedName].ExistingSecretReferences[secretObj.Name].SecretResourceVersion = secretObj.ResourceVersion
-		klog.V(5).Infof("Secret %q in %q namespace is %s", secretObj.Name, secretObj.Namespace, string(operationResult))
+			processor.Settings.SecretReferences[secretName].SecretResourceVersion = secretObj.ResourceVersion
+			klog.V(5).Infof("Secret %q in %q namespace is %s", secretObj.Name, secretObj.Namespace, string(operationResult))
+		} else {
+			if _, ok := reconciler.ProvidersReconcileState[namespacedName].ExistingSecretReferences[secretName]; ok {
+				processor.Settings.SecretReferences[secretName].SecretResourceVersion = reconciler.ProvidersReconcileState[namespacedName].ExistingSecretReferences[secretName].SecretResourceVersion
+			}
+			klog.V(5).Infof("Skip updating the secret %q in %q namespace since its data is up-to-date", secretName, provider.Namespace)
```
RichardChen820, line 463: Skip updating the secret %q in %q namespace since data is not changed.

internal/controller/appconfigurationprovider_controller.go
```diff
@@ -425,36 +425,43 @@ func (reconciler *AzureAppConfigurationProviderReconciler) createOrUpdateSecrets
...
-	for secretName, secret := range settings.SecretSettings {
-		secretObj := &corev1.Secret{
-			ObjectMeta: metav1.ObjectMeta{
-				Name:      secretName,
-				Namespace: provider.Namespace,
-			},
-			Type: secret.Type,
-		}
+	for secretName, secret := range processor.Settings.SecretSettings {
+		if processor.ShouldReconcile || shouldCreateOrUpdate(reconciler.ProvidersReconcileState[namespacedName].ExistingSecretReferences, secretName, processor.Settings.SecretReferences[secretName]) {
```
RichardChen820, line 429: This line is too long, break it into two lines.

internal/controller/processor.go
RichardChen820, line 286: nit: add empty line before and after this if block?

internal/controller/appconfigurationprovider_controller.go
RichardChen820, line 429: Nit: I know there's much difference in logic, but I think doing this is better for readability.

1. check `shouldReconcile` in the  `shouldCreateOrUpdate` method

2.
```
if !shouldCreateOrUpdate {
    // do something
  continue
}

// do something
```

internal/controller/appconfigurationprovider_controller_test.go
```diff
@@ -728,6 +763,7 @@ var _ = Describe("AppConfiguationProvider controller", func() {
 			Expect(string(secret.Data["testSecretKey3"])).Should(Equal("newTestSecretValue3"))
 			Expect(secret.Type).Should(Equal(corev1.SecretType("Opaque")))
 		})
+
```
RichardChen820, line 766: Nit: extra line

internal/controller/appconfigurationprovider_controller_test.go
```diff
@@ -713,7 +718,37 @@ var _ = Describe("AppConfiguationProvider controller", func() {
...
+				Type:           corev1.SecretType("Opaque"),
+				SecretMetadata: newSecretMetadata,
+			}
+
+			newTargetSettings := &loader.TargetKeyValueSettings{
+				SecretSettings:   newResolvedSecret,
+				SecretReferences: mockedSecretReference,
+			}
+
+			mockConfigurationSettings.EXPECT().ResolveSecretReferences(gomock.Any(), gomock.Any(), gomock.Any()).Return(newTargetSettings, nil)
```
RichardChen820, line 736: I didn't quite sure why check the refresh two times. And seems you didn't add the mocks for the secretId, how it can be refreshed without checking secretID?

internal/controller/processor.go
RichardChen820: Does `map.Clone()` help?

internal/controller/utils.go
```diff
@@ -302,3 +302,25 @@ func verifySelectorObject(selector acpv1.Selector) error {
 
 	return nil
 }
+
+func shouldCreateOrUpdate(existingSecretReferences map[string]*loader.TargetSecretReference, secretName string, secretReference *loader.TargetSecretReference) bool {
```
RichardChen820, line 306: Why not pass the existing and target secretMetadata as input?

internal/controller/appconfigurationprovider_controller_test.go
linglingye001: This test covers 2 refresh scenarios, the first time refresh when secretMetaData change, the second time not refresh when data not change. The first time refresh because mocked secretReferences length changed. Will add secretID.

internal/controller/utils.go
linglingye001: Target secretMetadata may not exist in existingSecretReferences map

internal/loader/configuration_setting_loader.go
```diff
@@ -50,7 +50,7 @@ type TargetKeyValueSettings struct {
 
 type TargetSecretReference struct {
 	Type                  corev1.SecretType
-	UriSegments           map[string]KeyVaultSecretUriSegment
+	SecretMetadata        map[string]KeyVaultSecretMetadata
```
RichardChen820, line 53: `SecretMetadatas`

linglingye001: `MetaData` is plural noun, no need `s`

internal/controller/appconfigurationprovider_controller_test.go
RichardChen820: Got you, and please add the scenario for the secretId check

internal/controller/processor.go
linglingye001: Have a try with `maps.Clone()` while it failed to apply to `maps[string]*loader.TargetSecretReference`, since the value type is pointer, after clone the new map points to the same struct.

internal/controller/appconfigurationprovider_controller.go
```diff
@@ -425,7 +425,15 @@ func (reconciler *AzureAppConfigurationProviderReconciler) createOrUpdateSecrets
 		Namespace: provider.Namespace,
 	}
 
-	for secretName, secret := range settings.SecretSettings {
+	for secretName, secret := range processor.Settings.SecretSettings {
+		if !shouldCreateOrUpdate(reconciler.ProvidersReconcileState[namespacedName].ExistingSecretReferences, secretName, processor.Settings.SecretReferences[secretName], processor.ShouldReconcile) {
```
RichardChen820, line 429: Just need to pass processor?

internal/controller/appconfigurationprovider_controller_test.go
```diff
@@ -713,7 +714,42 @@ var _ = Describe("AppConfiguationProvider controller", func() {
...
+			// Refresh interval is 1 minute, wait for 65 seconds to make sure the refresh is triggered
+			time.Sleep(65 * time.Second)
+
+			Eventually(func() bool {
+				err := k8sClient.Get(ctx, secretLookupKey, secret)
+				return err == nil
+			}, timeout, interval).Should(BeTrue())
+
+			Expect(secret.Namespace).Should(Equal(ProviderNamespace))
+			Expect(string(secret.Data["testSecretKey"])).Should(Equal("newTestSecretValue"))
```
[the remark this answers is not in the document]
linglingye001: line746-763 include scenario when `secretID` not changed.

internal/loader/configuration_setting_loader.go
```diff
@@ -245,7 +245,7 @@ func (csl *ConfigurationSettingLoader) CreateKeyValueSettings(ctx context.Contex
 			}
 
 			currentUrl := *setting.Value
-			secretUriSegment, err := parse(currentUrl)
+			secretMetadata, err := parse(currentUrl)
```
RichardChen820, line 248: This line and line200 are all using the secretMatadata word, that indeed causes confusion. I would suggest `SecretMatadatas`(regardless the plural and singular same form) or `SecretsMetadata` in `TargetSecretReference`

internal/controller/processor.go
```diff
@@ -201,17 +199,26 @@ func (processor *AppConfigurationProviderProcessor) processSecretReferenceRefres
 
 	// Only resolve the secret references that not specified the secret version
 	secretReferencesToSolve := make(map[string]*loader.TargetSecretReference)
+	copiedSecretReferences := make(map[string]*loader.TargetSecretReference)
 	for secretName, reference := range reconcileState.ExistingSecretReferences {
-		for key, uriSegment := range reference.UriSegments {
+		for key, uriSegment := range reference.SecretsMetadata {
```
RichardChen820, line 204: uriSegment -> secretMetadata

internal/controller/utils.go
```diff
@@ -302,3 +302,31 @@ func verifySelectorObject(selector acpv1.Selector) error {
...
+	secretReference := processor.Settings.SecretReferences[secretName]
+	if len(existingSecretReferences[secretName].SecretsMetadata) != len(secretReference.SecretsMetadata) {
+		return true
+	}
+
+	for key, secretMetadata := range secretReference.SecretsMetadata {
+		if _, ok := existingSecretReferences[secretName].SecretsMetadata[key]; !ok {
+			return true
+		}
+		if existingSecretReferences[secretName].SecretsMetadata[key].SecretId != nil && secretMetadata.SecretId != nil &&
```
RichardChen820, line 325: Break this long check into multiple lines

internal/loader/configuration_setting_loader.go
```diff
@@ -386,7 +387,10 @@ func (csl *ConfigurationSettingLoader) ResolveSecretReferences(
 						}
 						lock.Lock()
 						defer lock.Unlock()
-						resolvedSecretReferences[name].Data[currentKey] = []byte(*resolvedSecret.Value)
+						resolvedSecrets[name].Data[currentKey] = []byte(*resolvedSecret.Value)
+						currentUriSegment := targetSecretReference.SecretsMetadata[currentKey]
```
RichardChen820, line 391: Maybe I just didn't get what you are trying to do here, why don't just do:

```
targetSecretReference.SecretsMetadata[currentKey].SecretId= resolvedSecret.ID

```

linglingye001: In Golang, cannot assign to struct field in `map[string]*targetSecretReference`

RichardChen820: I see

internal/loader/configuration_setting_loader.go
```diff
@@ -386,7 +387,10 @@ func (csl *ConfigurationSettingLoader) ResolveSecretReferences(
 						}
 						lock.Lock()
 						defer lock.Unlock()
-						resolvedSecretReferences[name].Data[currentKey] = []byte(*resolvedSecret.Value)
+						resolvedSecrets[name].Data[currentKey] = []byte(*resolvedSecret.Value)
+						currentUriSegment := targetSecretReference.SecretMetadata[currentKey]
```
RichardChen820: This name is not accurate now,  currentSecretMetadata

RichardChen820 (approved):

PR #59, opened 2024-07-15
Refresh ConfigMap when data change

linglingye001:


internal/controller/appconfigurationprovider_controller.go
```diff
@@ -51,9 +51,10 @@ type AzureAppConfigurationProviderReconciler struct {
 type ReconciliationState struct {
 	Generation                              int64
 	ConfigMapResourceVersion                *string
-	SentinelETags                           map[acpv1.Sentinel]*azcore.ETag
```
RichardChen820, line 54: I thought the Sentinel Etags is still required to be noted, we still need to check etags of the sentinels.

linglingye001: Yes, if user set sentinels, they would be cached  in `KeyValueEtags`.

internal/controller/processor.go
```diff
@@ -288,8 +316,16 @@ func (processor *AppConfigurationProviderProcessor) Finish() (ctrl.Result, error
 		processor.ReconciliationState.ExistingSecretReferences = processor.Settings.SecretReferences
 	}
 
+	if len(processor.RefreshOptions.updatedKeyValueETags) > 0 {
```
RichardChen820, line 319: How abou tusing nil check instead of length check? If not updated, return nil.

internal/loader/settings_client.go
```diff
@@ -11,26 +11,78 @@ import (
...
+	"k8s.io/klog"
 )
 
 //go:generate mockgen -destination=mocks/mock_settings_client.go -package mocks . SettingsClient
 
-type SelectorSettingsClient struct {
-	selectors []acpv1.Selector
+type EtagSettingsClient struct {
+	etags       map[acpv1.Selector][]*azcore.ETag
+	etagChanged bool
```
RichardChen820, line 21: I don't expect to have this field in `EtagSettingsClient`, if the etag is not change, just return nil etgas is fine.

internal/loader/settings_client.go
```diff
@@ -11,26 +11,78 @@ import (
...
 	etag            *azcore.ETag
 	refreshInterval string
 }
+type SelectorSettingsClient struct {
+	selectors []acpv1.Selector
+}
 
 type SettingsClient interface {
-	GetSettings(ctx context.Context, client *azappconfig.Client) ([]azappconfig.Setting, error)
+	GetSettings(ctx context.Context, client *azappconfig.Client) ([]azappconfig.Setting, map[acpv1.Selector][]*azcore.ETag, error)
```
RichardChen820, line 34: I don't really the return type in an interface method, could we create a new type for them?

internal/loader/settings_client.go
```diff
@@ -11,26 +11,78 @@ import (
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/Azure/azure-sdk-for-go/sdk/data/azappconfig"
-	"k8s.io/klog/v2"
```
RichardChen820, line 14: Why do this change?

internal/loader/configuration_setting_loader.go
```diff
@@ -354,6 +331,56 @@ func (csl *ConfigurationSettingLoader) CheckAndRefreshSentinels(
 	return sentinelChanged, refreshedETags, nil
 }
 
+func (csl *ConfigurationSettingLoader) CheckPageETags(ctx context.Context, provider *acpv1.AzureAppConfigurationProvider, eTags map[acpv1.Selector][]*azcore.ETag) (bool, error) {
```
RichardChen820, line 334: there are AzureAppConfigurationProvider in the ConfigurationSettingLoader, don't need to pass it as a parameter

internal/loader/configuration_setting_loader.go
```diff
@@ -46,6 +46,8 @@ type TargetKeyValueSettings struct {
 	// Multiple secrets could be managed
 	SecretSettings   map[string]corev1.Secret
 	SecretReferences map[string]*TargetSecretReference
+	KeyValueETags    map[acpv1.Selector][]*azcore.ETag
```
RichardChen820, line 49: Can we avoid using both `KeyValueEtags` and `FeatureFlagETags`, can unify them into `PageEtags` ?

RichardChen820: Ok, we can't since it may return them both

Makefile
```diff
@@ -87,62 +127,72 @@ ifndef ignore-not-found
...
+deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
 	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
-	$(KUSTOMIZE) build config/default | kubectl apply -f -
+	$(KUSTOMIZE) build config/default | $(KUBECTL) apply -f -
 
 .PHONY: undeploy
-undeploy: ## Undeploy controller from the K8s cluster specified in ~/.kube/config. Call with ignore-not-found=true to ignore resource not found errors during deletion.
-	$(KUSTOMIZE) build config/default | kubectl delete --ignore-not-found=$(ignore-not-found) -f -
-
-.PHONY: helm-generate
```
RichardChen820, line 106: Let's keep these two items.

RichardChen820: I mean the helm-generate and helm-push

api/v1/azureappconfigurationprovider_types.go
```diff
@@ -53,7 +53,7 @@ type AzureAppConfigurationProviderStatus struct {
 // RefreshStatus defines last refresh time of configmap and secret when dynamic feature is enabled
 type RefreshStatus struct {
 	LastKeyVaultReferenceRefreshTime metav1.Time `json:"lastKeyVaultReferenceRefreshTime,omitempty"`
-	LastSentinelBasedRefreshTime     metav1.Time `json:"lastSentinelBasedRefreshTime,omitempty"`
+	LastKeyValueRefreshTime          metav1.Time `json:"lastKeyValueRefreshTime,omitempty"`
```
RichardChen820, line 56: Please test the upgrade scenario to see whether the field change in CRD would break upgrade scenario, You can try install 1.3.1 first then upgrade to this version to see if it block anything.

go.mod
```diff
@@ -82,6 +82,7 @@ require (
 	gopkg.in/yaml.v3 v3.0.1
 	k8s.io/api v0.30.1
 	k8s.io/apiextensions-apiserver v0.30.1 // indirect
+	k8s.io/klog v1.0.0
```
RichardChen820, line 85: Why we ned the klog v1?

internal/controller/processor.go
```diff
@@ -65,12 +68,15 @@ func (processor *AppConfigurationProviderProcessor) PopulateSettings(existingCon
 }
 
 func (processor *AppConfigurationProviderProcessor) processFullReconciliation() error {
+	processor.ReconciliationState.KeyValueETags, processor.ReconciliationState.FeatureFlagETags = initializeEtags(processor.Provider)
```
RichardChen820, line 71: What is the reason that `initializeEtags` is required?

internal/loader/settings_client.go
```diff
@@ -11,26 +11,78 @@ import (
...
 
 type SettingsClient interface {
-	GetSettings(ctx context.Context, client *azappconfig.Client) ([]azappconfig.Setting, error)
+	GetSettings(ctx context.Context, client *azappconfig.Client) ([]azappconfig.Setting, map[acpv1.Selector][]*azcore.ETag, error)
 }
 
-func (s *SentinelSettingsClient) GetSettings(ctx context.Context, client *azappconfig.Client) ([]azappconfig.Setting, error) {
+func (s *EtagSettingsClient) GetSettings(ctx context.Context, client *azappconfig.Client) ([]azappconfig.Setting, map[acpv1.Selector][]*azcore.ETag, error) {
+	nullString := "\x00"
+	for filter, pageEtags := range s.etags {
```
RichardChen820, line 39: Could we check the `etags` is nil or not? if it's nil, just need to return non-nil etags

linglingye001: Do you mean check `s.etags`? What scenario will it be `nil`?

config/crd/bases/azconfig.io_azureappconfigurationproviders.yaml
```diff
@@ -345,9 +347,3 @@ spec:
     storage: true
     subresources:
       status: {}
-status:
```
RichardChen820, line 348: I don't expect to remove this section

internal/loader/configuraiton_setting_loader_test.go
```diff
@@ -5,6 +5,7 @@ package loader
 
 import (
 	acpv1 "azappconfig/provider/api/v1"
+	v1 "azappconfig/provider/api/v1"
```
RichardChen820, line 8: Duplicate with line 7

internal/loader/settings_client.go
```diff
@@ -11,26 +11,78 @@ import (
...
+	etags       map[acpv1.Selector][]*azcore.ETag
+	etagChanged bool
 }
 
 type SentinelSettingsClient struct {
 	sentinel        acpv1.Sentinel
 	etag            *azcore.ETag
 	refreshInterval string
 }
+type SelectorSettingsClient struct {
```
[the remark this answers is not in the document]
RichardChen820: This one is not addressed

config/crd/bases/azconfig.io_azureappconfigurationproviders.yaml
linglingye001: It's auto-generated by make command. What's the influence of removing this section?

internal/loader/settings_client.go
```diff
@@ -47,19 +103,24 @@ func (s *SentinelSettingsClient) GetSettings(ctx context.Context, client *azappc
...
+	return &SettingsResponse{
+		Settings: []azappconfig.Setting{sentinelSetting.Setting},
+	}, nil
 }
 
-func (s *SelectorSettingsClient) GetSettings(ctx context.Context, client *azappconfig.Client) ([]azappconfig.Setting, error) {
-	settingsToReturn := make([]azappconfig.Setting, 0)
+func (s *SelectorSettingsClient) GetSettings(ctx context.Context, client *azappconfig.Client) (*SettingsResponse, error) {
 	nullString := "\x00"
+	settingsToReturn := make([]azappconfig.Setting, 0)
```
RichardChen820, line 122: `settings` is fine

internal/loader/settings_client.go
```diff
@@ -47,19 +103,24 @@ func (s *SentinelSettingsClient) GetSettings(ctx context.Context, client *azappc
...
+	refreshedEtags := make(map[acpv1.Selector][]*azcore.ETag)
```
RichardChen820, line 123: `pageEtags` is fine

internal/loader/settings_client.go
```diff
@@ -72,15 +133,19 @@ func (s *SelectorSettingsClient) GetSettings(ctx context.Context, client *azappc
 				Fields:      azappconfig.AllSettingFields(),
 			}
 			pager := client.NewListSettingsPager(selector, nil)
+			latestEtags := make([]*azcore.ETag, 0)
 
 			for pager.More() {
 				page, err := pager.NextPage(ctx)
 				if err != nil {
 					return nil, err
 				} else if len(page.Settings) > 0 {
```
RichardChen820, line 142: Would this check be problematic? What if it returns an empty setting list?

internal/loader/settings_client.go
```diff
@@ -16,21 +16,77 @@ import (
...
+			}
+
+			if pageCount != len(pageEtags) {
+				return settingsResponse, nil
+			}
+		}
+	}
+
+	// no change in the settings, return nil
+	return nil, nil
```
RichardChen820, line 86: `GetSetting()` should never return nil without any error, I would suggest return nil etags and settings rather than nil response.

internal/controller/processor.go
```diff
@@ -123,48 +138,60 @@ func (processor *AppConfigurationProviderProcessor) processKeyValueRefresh(exist
...
-		reconcileState.NextSentinelBasedRefreshReconcileTime = nextSentinelBasedRefreshReconcileTime
-		return nil
-	}
 	// Get the latest key value settings
 	existingConfigMapSettings := &existingConfigMap.Data
 	if processor.Settings.ConfigMapSettings != nil {
 		existingConfigMapSettings = &processor.Settings.ConfigMapSettings
 	}
+
+	if provider.Spec.Configuration.Refresh.Monitoring != nil {
```
RichardChen820, line 164: Could we put the line164- line178 right before the line 158, seems don't need to read the existing data if etag checking decides not refresh

internal/loader/settings_client.go
linglingye001: Yes, it should check if `page.Settings == nil`, when empty setting returned, `pageEtag` is not empty.

RichardChen820 (approved):

PR #60, opened 2024-07-18
Require to opt-in for the global service account

RichardChen820:
Require to opt in the global service account

### DIFF
--- a/deploy/parameter/helm-values.yaml
+++ b/deploy/parameter/helm-values.yaml
@@ -20,7 +20,7 @@ requestTracing:
 
 workloadIdentity:
   enabled: true
-  disableGlobalServiceAccount: false
+  enableGlobalServiceAccount: false
 
 serviceAccount:
   # Specifies whether a service account should be created
--- a/deploy/parameter/helm-values.yaml
+++ b/deploy/parameter/helm-values.yaml
@@ -20,7 +20,7 @@ requestTracing:
 
 workloadIdentity:
   enabled: true
-  enableGlobalServiceAccount: false
+  globalServiceAccountEnabled: false
 
 serviceAccount:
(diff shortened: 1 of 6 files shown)


deploy/parameter/helm-values.yaml
```diff
@@ -20,7 +20,7 @@ requestTracing:
 
 workloadIdentity:
   enabled: true
-  disableGlobalServiceAccount: false
+  enableGlobalServiceAccount: false
```
zhenlan, line 23: This is a verb. As a property name, it should be `globalServiceAccountEnabled`. Alternatively, you can have `useGlobalServiceAccount` or `globalServiceAccountAllowed`. Any preference?

RichardChen820: globalServiceAccountEnable is good to me

internal/controller/utils.go
```diff
@@ -247,14 +247,14 @@ func verifyWorkloadIdentityParameters(workloadIdentity *acpv1.WorkloadIdentityPa
 	var authCount int = 0
 
 	if workloadIdentity.ManagedIdentityClientId != nil {
-		if strings.EqualFold(os.Getenv(WorkloadIdentityDisableGlobalServiceAccount), "true") {
+		if strings.EqualFold(os.Getenv(WorkloadIdentityGlobalServiceAccountEnabled), "false") {
```
zhenlan, line 250: Why are we reading it from the environment? I thought it's a parameter passed in.

RichardChen820: The gap is how the controller can know the global service account is enabled or disabled, it just know it via the environment variable that been set while installation.

zhenlan (approved):

PR #62, opened 2024-07-26
Load balance mode

linglingye001:


internal/loader/configuration_setting_loader.go
```diff
@@ -480,6 +482,21 @@ func (csl *ConfigurationSettingLoader) ExecuteFailoverPolicy(ctx context.Context
...
+	if csl.AzureAppConfigurationProvider.Spec.LoadBalancingEnabled && csl.lastSuccessfulEndpoint != "" && len(clients) > 1 {
+		nextClientIndex := 0
+		for _, clientWrapper := range clients {
+			nextClientIndex++
+			if clientWrapper.Endpoint == csl.lastSuccessfulEndpoint {
+				break
+			}
+		}
+
+		// If we found the last successful client,we'll rotate the list so that the next client is at the beginning
```
zhiyuanliang-ms, line 494:
```suggestion
		// If we found the last successful client, we'll rotate the list so that the next client is at the beginning
```

internal/loader/configuration_setting_loader.go
```diff
@@ -508,6 +510,21 @@ func (csl *ConfigurationSettingLoader) ExecuteFailoverPolicy(ctx context.Context
...
+		if nextClientIndex < len(clients) {
+			clients = append(clients[nextClientIndex:], clients[:nextClientIndex]...)
```
RichardChen820, line 524: I have concern about this, this would be executed in super high frequency, almost every time of executing the `ExecuteFailoverPolicy()` right? It would create new array every time, have to create a new memory space to allocate the new array every time.

RichardChen820: We can have more effective to do the array rotation.

linglingye001: Array rotation is executed every time when `ExecuteFailoverPolicy` called, I don't think the frequency will be super high. It depends on the refresh interval the customer set and if load balance mode enabled. Besides, the number of replicas usually will not reach hundreds, O(n) time complexity is acceptable.

RichardChen820: But it may have many AzureAppConfigurationProvider yaml resources in one cluster to reconcile, and the improvement would be super easy to implement, so I think it's worth making a change.

PR #65, opened 2024-08-15
Load balance support

linglingye001:
Support load balance when `replicaDiscoveryEnabled`
Add load balance and failover request tracing


internal/loader/configuraiton_setting_loader_test.go
```diff
@@ -1169,6 +1169,156 @@ var _ = Describe("AppConfiguationProvider Get All Settings", func() {
...
+			Expect(len(allSettings.ConfigMapSettings)).Should(Equal(6))
+			Expect(allSettings.ConfigMapSettings["someKey1"]).Should(Equal("value1"))
+			Expect(allSettings.ConfigMapSettings["app:"]).Should(Equal("value2"))
+			Expect(allSettings.ConfigMapSettings["test:"]).Should(Equal("value3"))
+			Expect(allSettings.ConfigMapSettings["app:someSubKey1:1"]).Should(Equal("value4"))
+			Expect(allSettings.ConfigMapSettings["app:test:some"]).Should(Equal("value5"))
+			Expect(allSettings.ConfigMapSettings["app:test:"]).Should(Equal("value6"))
+		})
+
+		It("Succeed to get settings when load balance not enabled", func() {
```
RichardChen820, line 1248: Is the test necessary?

internal/loader/configuration_setting_loader.go
```diff
@@ -796,3 +830,26 @@ func MergeSecret(secret map[string]corev1.Secret, newSecret map[string]corev1.Se
 
 	return nil
 }
+
+// rotates the slice to the left by k positions
+func rotate(clients []*ConfigurationClientWrapper, k int) {
+	n := len(clients)
```
RichardChen820, line 836: check the `n` should not equels `0`

internal/loader/configuraiton_setting_loader_test.go
linglingye001: Yes, test only single client used (line 1311) when load balance not enabled.

internal/loader/configuration_setting_loader.go
```diff
@@ -535,8 +560,17 @@ func (csl *ConfigurationSettingLoader) ExecuteFailoverPolicy(ctx context.Context
 func updateClientBackoffStatus(clientWrapper *ConfigurationClientWrapper, successful bool) {
 	if successful {
 		clientWrapper.BackOffEndTime = metav1.Time{}
-		clientWrapper.FailedAttempts = 0
+		// Reset FailedAttempts when client succeeded
+		if clientWrapper.FailedAttempts > 0 {
+			clientWrapper.FailedAttempts = 0
+		}
+		// Use negative value to indicate that successful attempt
+		clientWrapper.FailedAttempts--
```
RichardChen820, line 568: Set `clientWrapper.FailedAttempts = -1`  directly?

RichardChen820: Why not use `0` to indicate success?

linglingye001: `0` can also represent the client is not used. negative value can make sure the client's status update when it succeeded.

internal/loader/request_tracing.go
```diff
@@ -1,73 +1,81 @@
...
+			if manager.IsFailoverRequest {
+				output = append(output, "FailoverRequest")
+			}
+		}
+	}
+
+	if provider.Spec.LoadBalancingEnabled {
+		output = append(output, "UsesLoadBalancing")
+	}
+
```
RichardChen820, line 71: Please align the format with .net provider

internal/loader/configuration_setting_loader.go
```diff
@@ -502,27 +504,50 @@ func (csl *ConfigurationSettingLoader) ExecuteFailoverPolicy(ctx context.Context
...
 
 	errors := make([]error, 0)
+	var tracingEnabled bool
+	if value, ok := os.LookupEnv(RequestTracingEnabled); ok {
+		tracingEnabled, _ = strconv.ParseBool(value)
+	}
 	for _, clientWrapper := range clients {
-		successful := true
+		if tracingEnabled {
+			ctx = policy.WithHTTPHeader(ctx, createCorrelationContextHeader(ctx, csl.AzureAppConfigurationProvider, csl.ClientManager))
```
RichardChen820, line 529: This seems problematic, it traverse the clients , would accumulates the traces values to the same `ctx` object.

RichardChen820: Forget it, it won't be a problem.

internal/loader/configuration_setting_loader.go
```diff
@@ -502,27 +502,51 @@ func (csl *ConfigurationSettingLoader) ExecuteFailoverPolicy(ctx context.Context
...
+		}
 		settingsResponse, err := settingsClient.GetSettings(ctx, clientWrapper.Client)
+		successful := true
 		if err != nil {
 			successful = false
 			updateClientBackoffStatus(clientWrapper, successful)
 			if IsFailoverable(err) {
 				klog.Warningf("current client of '%s' failed to get settings: %s", clientWrapper.Endpoint, err.Error())
 				errors = append(errors, err)
+				if ok {
```
RichardChen820, line 538: `ok` here just means it can find the RequestTracingEnabled in env var.

internal/loader/configuration_client_manager.go
```diff
@@ -37,6 +37,8 @@ import (
 
 type ConfigurationClientManager struct {
 	ReplicaDiscoveryEnabled   bool
+	LoadBalancingEnabled      bool
+	IsFailoverRequest         bool
```
RichardChen820, line 41: I don't think it's required to be a property in `ConfigurationClientManager`, we don't need to remember the state of `IsFailoverRequest`

RichardChen820 (approved):

PR #70, opened 2024-09-10
Set label filter to nil when it's empty string

linglingye001:


internal/loader/configuration_setting_loader.go
```diff
@@ -664,7 +664,15 @@ func deduplicateFilters(filters []acpv1.Selector) []acpv1.Selector {
 				}
 			}
 			if !findDuplicate {
-				result = append(result, filters[i])
+				labelFilter := filters[i].LabelFilter
```
RichardChen820, line 667: I don't think we should do the conversion while deduplicating the selectors, we need to do it more straightforwardly.

RichardChen820: And what about the label for the `sentinel`?

api/v1/azureappconfigurationprovider_types.go
```diff
@@ -106,9 +106,9 @@ type RefreshMonitoring struct {
 
 // Defines the keyValues to be watched.
 type Sentinel struct {
-	Key string `json:"key"`
+	Key *string `json:"key"`
```
RichardChen820, line 109: Key should not be nullable.

internal/loader/configuration_setting_loader.go
```diff
@@ -643,6 +644,40 @@ func GetFeatureFlagFilters(acpSpec acpv1.AzureAppConfigurationProviderSpec) []ac
...
+		results = append(results, acpv1.Sentinel{
+			Key:   sentinel.Key,
+			Label: label,
+		})
+
+	}
+	return results
+}
+
+func processEmptyLabelFilter(filters []acpv1.Selector) []acpv1.Selector {
```
RichardChen820, line 664: How about rename to "normalizeLableFilter"?

internal/loader/settings_client.go
```diff
@@ -88,15 +88,15 @@ func (s *EtagSettingsClient) GetSettings(ctx context.Context, client *azappconfi
 }
 
 func (s *SentinelSettingsClient) GetSettings(ctx context.Context, client *azappconfig.Client) (*SettingsResponse, error) {
-	sentinelSetting, err := client.GetSetting(ctx, s.sentinel.Key, &azappconfig.GetSettingOptions{Label: &s.sentinel.Label, OnlyIfChanged: s.etag})
+	sentinelSetting, err := client.GetSetting(ctx, s.sentinel.Key, &azappconfig.GetSettingOptions{Label: s.sentinel.Label, OnlyIfChanged: s.etag})
```
RichardChen820, line 91: We need to do the same as  `SelectorSettingsClient`,  if label is nil, set it to `\x00`

RichardChen820 (approved):

PR #74, opened 2024-10-21
Feature flag telemetry metadata support

linglingye001:


internal/loader/configuration_setting_loader.go
```diff
@@ -894,3 +918,184 @@ func reverseClients(clients []*ConfigurationClientWrapper, start, end int) {
 		end--
 	}
 }
+
+func calculateFeatureFlagId(setting azappconfig.Setting) string {
+	// Create the basic value string
+	featureFlagValue := *setting.Key + "\n"
```
RichardChen820, line 924: It's not the value of featureFlag

internal/loader/configuration_setting_loader.go
```diff
@@ -894,3 +918,184 @@ func reverseClients(clients []*ConfigurationClientWrapper, start, end int) {
...
+	// Generate SHA-256 hash, and encode it to Base64
+	hash := sha256.Sum256([]byte(featureFlagValue))
+	encodedFeatureFlag := base64.StdEncoding.EncodeToString(hash[:])
+
+	// Replace '+' with '-' and '/' with '_'
+	encodedFeatureFlag = strings.ReplaceAll(encodedFeatureFlag, "+", "-")
+	encodedFeatureFlag = strings.ReplaceAll(encodedFeatureFlag, "/", "_")
+
+	// Remove all instances of "=" at the end of the string that were added as padding
+	if idx := strings.Index(encodedFeatureFlag, "="); idx != -1 {
```
RichardChen820, line 938: Use the `strings.Trim()` instead?

internal/loader/configuration_setting_loader.go
```diff
@@ -894,3 +918,184 @@ func reverseClients(clients []*ConfigurationClientWrapper, start, end int) {
...
+			metadata, _ := telemetry[FeatureFlagMetadataKey].(map[string]interface{})
+			if metadata == nil {
+				metadata = make(map[string]interface{})
+			}
+
+			// Set the new metadata
+			metadata[FeatureFlagETagKey] = *setting.ETag
+			metadata[FeatureFlagIdKey] = calculateFeatureFlagId(setting)
+			metadata[FeatureFlagReferenceKey] = createFeatureFlagReference(setting, endpoint)
+			if allocationId := generateAllocationId(featureFlag); allocationId != "" {
```
RichardChen820, line 1062: In what case the allocationId could be empty?

internal/loader/configuration_setting_loader.go
```diff
@@ -894,3 +918,184 @@ func reverseClients(clients []*ConfigurationClientWrapper, start, end int) {
...
+
+	// Convert to base64 URL-safe string
+	allocationId := base64.RawURLEncoding.EncodeToString(first15Bytes)
+	return allocationId
+}
+
+func updateFeatureFlagTelemetry(featureFlag map[string]interface{}, setting azappconfig.Setting, endpoint string) {
+	if telemetry, ok := featureFlag[FeatureFlagTelemetryKey].(map[string]interface{}); ok {
+		if enabled, ok := telemetry[FeatureFlagEnabledKey].(bool); ok && enabled {
+			metadata, _ := telemetry[FeatureFlagMetadataKey].(map[string]interface{})
```
RichardChen820, line 1053: In what case the metadata field already exists here?

internal/loader/configuration_setting_loader.go
```diff
@@ -894,3 +918,184 @@ func reverseClients(clients []*ConfigurationClientWrapper, start, end int) {
...
+	}
+}
+
+func contains(arr []string, target string) bool {
+	for _, a := range arr {
+		if a == target {
+			return true
+		}
+	}
+	return false
```
RichardChen820, line 1076: nit: add an empty line before return

internal/loader/configuration_setting_loader.go
```diff
@@ -898,3 +923,184 @@ func reverseClients(clients []*ConfigurationClientWrapper, start, end int) {
...
+
+	// Remove all instances of "=" at the end of the string that were added as padding
+	if idx := strings.Index(encodedFeatureFlag, "="); idx != -1 {
+		encodedFeatureFlag = encodedFeatureFlag[:idx]
+	}
+
+	return encodedFeatureFlag
+}
+
+func createFeatureFlagReference(setting azappconfig.Setting, endpoint string) string {
```
RichardChen820, line 950: `generateFeatureFlagReference` is better.

internal/loader/configuration_setting_loader.go
```diff
@@ -91,6 +94,23 @@ const (
 	FeatureFlagKeyPrefix                  string = ".appconfig.featureflag/"
 	FeatureFlagSectionName                string = "feature_flags"
 	FeatureManagementSectionName          string = "feature_management"
+	FeatureFlagTelemetryKey               string = "telemetry"
+	FeatureFlagEnabledKey                 string = "enabled"
+	FeatureFlagMetadataKey                string = "metadata"
+	FeatureFlagETagKey                    string = "ETag"
+	FeatureFlagIdKey                      string = "FeatureFlagId"
+	FeatureFlagReferenceKey               string = "FeatureFlagReference"
```
RichardChen820, line 102: Why these three three name with Pascal-casing that is different from others?

internal/loader/configuration_setting_loader.go
linglingye001: It's by design

internal/loader/configuration_setting_loader.go
linglingye001: It's by design

RichardChen820: Okay, Then I would suggest to properly group these keys, to indicate which are key for "telemetry only"

internal/loader/configuration_setting_loader.go
```diff
@@ -898,3 +923,184 @@ func reverseClients(clients []*ConfigurationClientWrapper, start, end int) {
...
+	if setting.Label != nil && strings.TrimSpace(*setting.Label) != "" {
+		// Encode the label to ensure it is safe for URLs
+		encodedLabel := url.QueryEscape(*setting.Label)
+		featureFlagReference += fmt.Sprintf("?label=%s", encodedLabel)
+	}
+
+	return featureFlagReference
+}
+
+func generateAllocationId(featureFlag map[string]interface{}) string {
```
RichardChen820, line 963: caculateAollcationId

internal/loader/configuration_setting_loader.go
```diff
@@ -898,3 +923,184 @@ func reverseClients(clients []*ConfigurationClientWrapper, start, end int) {
...
+	// Hash the raw allocation ID
+	hash := sha256.Sum256([]byte(rawAllocationId.String()))
+	first15Bytes := hash[:15]
+
+	// Convert to base64 URL-safe string
+	allocationId := base64.RawURLEncoding.EncodeToString(first15Bytes)
+	return allocationId
+}
+
+func updateFeatureFlagTelemetry(featureFlag map[string]interface{}, setting azappconfig.Setting, endpoint string) {
```
RichardChen820, line 1055: Is `populateTelemetryMetadata` better for the name of this method?

internal/loader/configuration_setting_loader.go
```diff
@@ -898,3 +923,184 @@ func reverseClients(clients []*ConfigurationClientWrapper, start, end int) {
...
+
+func calculateFeatureFlagId(setting azappconfig.Setting) string {
+	// Create the basic value string
+	featureFlagValue := *setting.Key + "\n"
+	if setting.Label != nil && strings.TrimSpace(*setting.Label) != "" {
+		featureFlagValue += *setting.Label
+	}
+
+	// Generate SHA-256 hash, and encode it to Base64
+	hash := sha256.Sum256([]byte(featureFlagValue))
```
RichardChen820, line 935: Could you please confirm this string -> bytes conversion is using the  UTF8 encode?

internal/loader/configuration_setting_loader.go
```diff
@@ -898,3 +923,184 @@ func reverseClients(clients []*ConfigurationClientWrapper, start, end int) {
...
+	}
+
+	return featureFlagReference
+}
+
+func generateAllocationId(featureFlag map[string]interface{}) string {
+	var rawAllocationId strings.Builder
+	var variantsForExperimentation []string
+
+	allocationSection, ok := featureFlag[AllocationKey].(map[string]interface{})
```
RichardChen820, line 967: `allocation` is fine

internal/loader/configuration_setting_loader.go
```diff
@@ -898,3 +923,184 @@ func reverseClients(clients []*ConfigurationClientWrapper, start, end int) {
...
+	if !ok {
+		return ""
+	}
+
+	// Seed
+	seedValue := ""
+	if allocationSection[SeedKey] != nil {
+		seedValue = allocationSection[SeedKey].(string)
+	}
+	rawAllocationId.WriteString(fmt.Sprintf("seed=%v\ndefault_when_enabled=", seedValue))
```
RichardChen820, line 977: Shouldn't be `rawAllocationId.WriteString(fmt.Sprintf("seed=%v", seedValue))`? `default_when_enabled=` should be added along with its value

internal/loader/configuration_setting_loader.go
```diff
@@ -898,3 +923,184 @@ func reverseClients(clients []*ConfigurationClientWrapper, start, end int) {
...
+
+	// Default variant when enabled
+	if defaultVariant, exists := allocationSection[DefaultWhenEnabledKey]; exists {
+		variantsForExperimentation = append(variantsForExperimentation, defaultVariant.(string))
```
RichardChen820, line 981: variantsForExperimentation -> allocatedVariants

internal/loader/configuration_setting_loader.go
```diff
@@ -898,3 +923,184 @@ func reverseClients(clients []*ConfigurationClientWrapper, start, end int) {
...
+		rawAllocationId.WriteString(fmt.Sprintf("%v", defaultVariant.(string)))
```
RichardChen820, line 982: `default_when_enabled= ` should be added here

internal/loader/configuration_setting_loader.go
```diff
@@ -898,3 +923,184 @@ func reverseClients(clients []*ConfigurationClientWrapper, start, end int) {
...
+	}
+
+	// Percentiles
+	rawAllocationId.WriteString("\npercentiles=")
```
RichardChen820, line 986: Could you confirm whether it's by design that keep `percentiles=` with no value while no percentile set?

internal/loader/configuration_setting_loader.go
```diff
@@ -898,3 +923,184 @@ func reverseClients(clients []*ConfigurationClientWrapper, start, end int) {
...
+			}
+		}
+	}
+
+	// Short-circuit if no valid data
+	if len(variantsForExperimentation) == 0 && allocationSection[SeedKey] == nil {
+		return ""
+	}
+
+	rawAllocationId.WriteString("\nvariants=")
```
RichardChen820, line 1019: Same with `percentile`

internal/loader/configuration_setting_loader.go
```diff
@@ -898,3 +923,184 @@ func reverseClients(clients []*ConfigurationClientWrapper, start, end int) {
...
+	return encodedFeatureFlag
+}
+
+func createFeatureFlagReference(setting azappconfig.Setting, endpoint string) string {
+	featureFlagReference := fmt.Sprintf("%s/kv/%s", endpoint, *setting.Key)
+
+	// Check if the label is present and not empty
+	if setting.Label != nil && strings.TrimSpace(*setting.Label) != "" {
+		// Encode the label to ensure it is safe for URLs
+		encodedLabel := url.QueryEscape(*setting.Label)
```
RichardChen820, line 956: Could confirm that in other language, they also do the URL encode?

internal/loader/configuration_setting_loader.go
```diff
@@ -898,3 +923,184 @@ func reverseClients(clients []*ConfigurationClientWrapper, start, end int) {
...
+		featureFlagReference += fmt.Sprintf("?label=%s", encodedLabel)
+	}
+
+	return featureFlagReference
+}
+
+func generateAllocationId(featureFlag map[string]interface{}) string {
+	var rawAllocationId strings.Builder
+	var variantsForExperimentation []string
```
RichardChen820, line 965: allocatedVariants

internal/loader/configuration_setting_loader.go
```diff
@@ -898,3 +923,184 @@ func reverseClients(clients []*ConfigurationClientWrapper, start, end int) {
...
+		}
+
+		sort.Slice(sortedPercentiles, func(i, j int) bool {
+			return sortedPercentiles[i][FromKey].(float64) < sortedPercentiles[j][FromKey].(float64)
+		})
+
+		for i, percentile := range sortedPercentiles {
+			variantsForExperimentation = append(variantsForExperimentation, percentile[VariantKey].(string))
+			rawAllocationId.WriteString(fmt.Sprintf("%v,%s,%v", percentile[FromKey].(float64), base64.StdEncoding.EncodeToString([]byte(percentile[VariantKey].(string))), percentile[ToKey].(float64)))
+			if i != len(sortedPercentiles)-1 {
```
RichardChen820, line 1008: Does golang have method like `string.Join(";")`?

internal/loader/configuration_setting_loader.go
```diff
@@ -898,3 +923,184 @@ func reverseClients(clients []*ConfigurationClientWrapper, start, end int) {
...
+		}
+
+		sort.Slice(sortedVariants, func(i, j int) bool {
+			return sortedVariants[i][NameKey].(string) < sortedVariants[j][NameKey].(string)
+		})
+
+		for i, variant := range sortedVariants {
+			configurationValue := []byte("")
+			if _, exist := variant[ConfigurationValueKey]; exist {
+				configurationValue, _ = json.Marshal(jsonSorter(variant[ConfigurationValueKey]))
```
RichardChen820, line 1038: What the configurationValue will be if json.Marshal a nil value?

internal/loader/configuration_setting_loader.go
```diff
@@ -898,3 +923,184 @@ func reverseClients(clients []*ConfigurationClientWrapper, start, end int) {
...
+func contains(arr []string, target string) bool {
+	for _, a := range arr {
+		if a == target {
+			return true
+		}
+	}
+	return false
+}
+
+func jsonSorter(v interface{}) interface{} {
```
RichardChen820, line 1084: `jsonSorter` sounds too odd,  could we let the method do the both sort and Serialization?

internal/loader/configuration_setting_loader.go
linglingye001: see https://github.com/Azure/AppConfiguration-DotnetProvider/pull/517#discussion_r1488671663 and https://datatracker.ietf.org/doc/html/rfc7515#appendix-C

internal/loader/configuration_setting_loader.go
linglingye001: No, we should not encode label.

internal/loader/configuration_setting_loader.go
linglingye001: `default_when_enabled=` can be added separately with `seed`

internal/loader/configuration_setting_loader.go
linglingye001: `exists` can be false.

internal/loader/configuration_setting_loader.go
linglingye001: It's by design.

internal/loader/configuration_setting_loader.go
linglingye001: Go has `Join(elems []string, sep string) string` for slice scenario

internal/loader/configuration_setting_loader.go
linglingye001: `jsonSorter` is not necessary since `json.Marshal` will sort map keys internally, https://pkg.go.dev/encoding/json#Marshal

internal/loader/configuration_setting_loader.go
linglingye001: Based on https://go.dev/blog/strings#TOC_2, Go source code is UTF-8, so the source code for the string literal is UTF-8 text. When constructed from string literals, those bytes are UTF-8

internal/loader/configuration_setting_loader.go
linglingye001: Not sure whether server side updates `telemetry.metadata`. The current implementation is consistent with other providers.

RichardChen820 (approved):

PR #77, opened 2024-10-31
Fix sentinel refresh and nil pointer issue

linglingye001:


internal/controller/processor.go
```diff
@@ -76,6 +76,7 @@ func (processor *AppConfigurationProviderProcessor) processFullReconciliation()
 	processor.RefreshOptions.ConfigMapSettingPopulated = true
 	processor.RefreshOptions.updatedKeyValueETags = updatedSettings.KeyValueETags
 	processor.RefreshOptions.updatedFeatureFlagETags = updatedSettings.FeatureFlagETags
+	processor.RefreshOptions.updatedSentinelETags = updatedSettings.SentinelETags
```
RichardChen820, line 79: Shall we check if the returned `updatedSettings.SentinelETags` is not nil, then we also change the `processor.RefreshOptions.sentinelChanged=true`? Then we don't have to add additional check in `Finish()`

RichardChen820: And we can also ensure that the `processor.ReconciliationState.SentinelETags` would never be set to nil

internal/loader/configuration_setting_loader.go
```diff
@@ -299,22 +315,14 @@ func (csl *ConfigurationSettingLoader) CheckAndRefreshSentinels(
...
-	}
 	refreshedETags := make(map[acpv1.Sentinel]*azcore.ETag)
-	sentinels := normalizeSentinels(provider.Spec.Configuration.Refresh.Monitoring.Sentinels)
-
-	for _, sentinel := range sentinels {
-		if eTag, ok := eTags[sentinel]; ok {
-			// Initialize the updatedETags with the current eTags
-			refreshedETags[sentinel] = eTag
-		}
+	for sentinel, currentETag := range eTags {
```
RichardChen820, line 319: Do we need to add a nil check for the `eTags`?

internal/controller/processor.go
linglingye001: There may be 2 scenarios when shouldReconcile is true. User adds monitoring section or removes it. If no sentinel is setup, we need to make sure the `ReconciliationState.SentinelETags` be updated to empty list.

linglingye001: I agree that sentinelETags should not be nil.

RichardChen820 (approved):

PR #97, opened 2025-03-05
Fix nil pointer issue when restart k8s provider

linglingye001:


internal/controller/processor.go
```diff
@@ -297,7 +297,8 @@ func (processor *AppConfigurationProviderProcessor) shouldReconcile(
 	}
 
 	for name, secret := range existingSecrets {
-		if processor.ReconciliationState.ExistingK8sSecrets[name].SecretResourceVersion != secret.ResourceVersion {
+		if processor.ReconciliationState.ExistingK8sSecrets[name] != nil &&
```
RichardChen820, line 300: The `processor.ReconciliationState.ExistingK8sSecrets` wouldn't be nil, right?

RichardChen820: Double checked, it can't be nil.

RichardChen820:
We need to some update to this logic as well, we need ensure the existing secrets are populated to the reconcileState.existingSecrets as well.

RichardChen820 (approved):

PR #110, opened 2025-04-17
AI Config Telemetry

linglingye001:


internal/loader/configuration_setting_loader.go
```diff
@@ -302,9 +311,22 @@ func (csl *ConfigurationSettingLoader) CreateKeyValueSettings(ctx context.Contex
 		default:
 			rawSettings.KeyValueSettings[trimmedKey] = setting.Value
 			rawSettings.IsJsonContentTypeMap[trimmedKey] = isJsonContentType(setting.ContentType)
+			if rawSettings.IsJsonContentTypeMap[trimmedKey] {
+				// if the content type is json, settings.ContentType is not nil
+				if strings.Contains(*setting.ContentType, AIMimeProfileKey) {
```
zhiyuanliang-ms, line 316: corner case: 

`setting.ContentType` is `application/json; foo="https://azconfig.io/mime-profiles/ai/chat-completion"`

internal/loader/configuration_setting_loader.go
```diff
@@ -109,11 +110,18 @@ const (
 	FeatureFlagReferenceKey string = "FeatureFlagReference"
 )
 
+// AI Configuration telemetry
+const (
+	AIMimeProfileKey               string = "https://azconfig.io/mime-profiles/ai"
+	AIChatCompletionMimeProfileKey string = "https://azconfig.io/mime-profiles/ai/chat-completion"
+)
+
```
RichardChen820, line 118: Nit: how about moving these constants into `request_tracing.go` file along with the feature flag telemetry constants abvoe,  since they are only telemetry related.

internal/loader/configuration_setting_loader.go
zhiyuanliang-ms: It is even better to make it case-insensitive.

internal/loader/request_tracing_test.go
```diff
@@ -0,0 +1,176 @@
...
+		},
+		{
+			name: "with features",
+			provider: acpv1.AzureAppConfigurationProvider{
+				Spec: acpv1.AzureAppConfigurationProviderSpec{
+					LoadBalancingEnabled: true,
+				},
+			},
+			tracingFeatures: TracingFeatures{
+				UseAIConfiguration:               true,
```
zhiyuanliang-ms, line 113: We need testcase to verify that if a configuration setting's content type has ai configuration profile, the tracingFeatures will be set correctly.

This unit test is not enough to make sure request tracing can work as expected.

internal/loader/request_tracing.go
```diff
@@ -51,22 +59,31 @@ func createCorrelationContextHeader(ctx context.Context, provider acpv1.AzureApp
...
+		if tracingFeatures.UseAIConfiguration {
+			features = append(features, "AI")
+		}
+
+		if tracingFeatures.UseAIChatCompletionConfiguration {
+			features = append(features, "AICC")
+		}
+
+		featureStr := "Features=" + strings.Join(features, TracingFeatureDelimiterKey)
+		output = append(output, featureStr)
```
RichardChen820, line 86: features := make([]string, 0)
                
                if provider.Spec.LoadBalancingEnabled {
			features = append(features, "LB")
		}

		if tracingFeatures.UseAIConfiguration {
			features = append(features, "AI")
		}

		if tracingFeatures.UseAIChatCompletionConfiguration {
			features = append(features, "AICC")
		}

              if len(features) >0 {
                   featureStr := "Features=" + strings.Join(features, TracingFeatureDelimiterKey)
		    output = append(output, featureStr)
                }

internal/loader/request_tracing.go
```diff
@@ -51,22 +59,31 @@ func createCorrelationContextHeader(ctx context.Context, provider acpv1.AzureApp
...
+		if provider.Spec.LoadBalancingEnabled {
+			features = append(features, "LB")
+		}
+
+		if tracingFeatures.UseAIConfiguration {
+			features = append(features, "AI")
+		}
+
+		if tracingFeatures.UseAIChatCompletionConfiguration {
+			features = append(features, "AICC")
```
RichardChen820, line 82: Nit:  use constant for `LB`, `AI`, `AICC`

internal/loader/request_tracing_test.go
linglingye001: tests added

zhiyuanliang-ms: Did you forgot to push the commit?

zhiyuanliang-ms:
According to the latest discussion, ff and key vault reference should be excluded. See this [PR](https://github.com/Azure/AppConfiguration-DotnetProvider/pull/651)

RichardChen820 (approved):

internal/loader/request_tracing_test.go
zhiyuanliang-ms: > We need testcase to verify that if the content type of a configuration setting from server has ai configuration profile, the tracingFeatures will be set correctly.

Which test did you add for this?

internal/loader/typed_setting_test.go
```diff
@@ -633,3 +633,131 @@ func TestBuildHierarchicalSetting(t *testing.T) {
...
 }
+
+func TestIsAIConfigurationContentType(t *testing.T) {
+	tests := []struct {
+		name        string
+		contentType *string
+		expected    bool
+	}{
+		{
+			name:        "valid AI configuration content type",
```
zhiyuanliang-ms, line 644: How about using `mock_configuration_settings_retriever` to mock a server response and build an end-to-end testcase? It will make us more confident about the code.

PR #164, opened 2026-03-11
Audience parameter support

linglingye001:
Usage
```
helm install azureappconfiguration.kubernetesprovider \
    oci://mcr.microsoft.com/azure-app-configuration/helmchart/kubernetes-provider \
    --namespace azappconfig-system \
    --create-namespace
    --set env.azureAppConfigurationAudience="https://appconfig.sovcloud-api.fr"
```

### DIFF
--- a/deploy/parameter/helm-values.yaml
+++ b/deploy/parameter/helm-values.yaml
@@ -67,6 +67,8 @@ autoscaling:
   targetCPUUtilizationPercentage: 80
   targetMemoryUtilizationPercentage: 80
 
+audience: ""
+
 env:
   azureClientId: ""
   azureTenantId: ""
--- a/deploy/parameter/helm-values.yaml
+++ b/deploy/parameter/helm-values.yaml
@@ -67,11 +67,10 @@ autoscaling:
   targetCPUUtilizationPercentage: 80
   targetMemoryUtilizationPercentage: 80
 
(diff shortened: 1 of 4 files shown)


deploy/parameter/helm-values.yaml
```diff
@@ -67,6 +67,8 @@ autoscaling:
   targetCPUUtilizationPercentage: 80
   targetMemoryUtilizationPercentage: 80
 
+audience: ""
```
RichardChen820: How about put it into `env` section and name it "azureAppconfigAudience"

linglingye001: We can use full product name. If put it to the `env` section , the command will be `--set env.azureAppConfigurationAudience`, what do you think? @zhenlan

zhenlan: Which one is more consistent with existing naming conventions?

linglingye001: Helm [best practice ](https://helm.sh/docs/chart_best_practices/values/) recommended camelCase, and `azureAppConfigurationAudience` is more descriptive, consistent with how `azureClientId` and `azureTenantId` are already done.

RichardChen820 (approved):

RichardChen820:
LGTM

PR #166, opened 2026-03-20
Snapshot reference support

linglingye001:


internal/loader/configuration_setting_loader.go
```diff
@@ -331,6 +347,136 @@ func (csl *ConfigurationSettingLoader) CreateKeyValueSettings(ctx context.Contex
...
+					if isAIConfigurationContentType(setting.ContentType) {
+						*useAIConfiguration = true
+					}
+					if isAIChatCompletionContentType(setting.ContentType) {
+						*useAIChatCompletionConfiguration = true
+					}
+				}
+			}
+		}
+	}
```
RichardChen820: We have the same `switch..case...` logic in the `CreateKeyValueSettings()` function, do we? Can we just make it as a function, and reuse in two place

linglingye001: updated

RichardChen820 (approved):

internal/loader/configuration_setting_loader.go
RichardChen820: approved

Issue #180, opened 2026-05-28
Snapshot reference auto-resolution silently overrides other selectors in the same `selectors` array


hereicq:
## Expected behavior

Per the [[reference doc](https://learn.microsoft.com/en-us/azure/azure-app-configuration/reference-kubernetes-provider#key-value-selection)](https://learn.microsoft.com/en-us/azure/azure-app-configuration/reference-kubernetes-provider#key-value-selection): *"the key-values of the last selector take precedence and override any overlapping keys from the previous selectors."*

The doc explicitly demonstrates this with a `snapshotName` selector followed by a `keyFilter` selector that overrides some of the snapshot's values:

```yaml
configuration:
  selectors:
    - snapshotName: app1_common_configuration
    - keyFilter: app1*
      labelFilter: development
```

> *"In the following sample, you load key-values of common configuration from a snapshot and then override some of them with key-values for development."*

A reasonable user expectation, given the v2.6.0 introduction of snapshot **references** (auto-resolved via the `snapshot-ref` content type), is that snapshot references participate in the chain identically to `snapshotName` selectors — so a later selector can override values pulled in via a reference.

## Actual behavior

A snapshot reference resolved via `configuration.selectors` **always wins** over any other selector in the chain that returns the same App Config key, regardless of:

- Selector order (placing the snapshot-ref selector first vs. last produces the same result).
- Override pattern (same-key in live store + co-located key-filter selector).
- Prefix-trim pattern (`live/<key>` in live store + `trimKeyPrefixes: ["live/"]` on the CR).

The override is silently lost — no warning, status stays `Complete`.

## Repro summary

Created an App Configuration store containing:
- `operational/foo = "baseline"` (label `test`)
- `.appconfig.featureflag/PocFeatureA` (label `test`)
- A snapshot capturing both above
- `system/active-snapshot` (label `test`) — snapshot-ref content-type, points at the snapshot
- Overwrote `operational/foo = "override-A"` in the live store *after* snapshot creation (snapshot's frozen value still `"baseline"`)

Applied the following CR:

```yaml
configuration:
  selectors:
    - keyFilter: "system/active-snapshot"   # 1st: snapshot reference, resolves to baseline
      labelFilter: test
    - keyFilter: "operational/*"             # 2nd: direct read from live store, returns override-A
      labelFilter: test
```

**Expected ConfigMap** (per the doc's last-wins contract): `operational/foo = "override-A"`.

[...]

PR #182, opened 2026-06-03
Fix snapshot reference resolution order

linglingye001:
Fix https://github.com/Azure/AppConfiguration-KubernetesProvider/issues/180

### DIFF
--- a/internal/loader/configuration_setting_loader.go
+++ b/internal/loader/configuration_setting_loader.go
@@ -208,7 +208,8 @@ func (csl *ConfigurationSettingLoader) RefreshFeatureFlagSettings(ctx context.Co
 type settingProcessContext struct {
 	rawSettings                      *RawSettings
 	resolver                         *SecretReferenceResolver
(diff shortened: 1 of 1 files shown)


zhenlan:
> * Introduce `allowSnapshotRef` to prevent nested snapshot reference resolution and `snapshotClient` to lazily initialize/reuse an App Configuration client for snapshot reads.

@linglingye001 does it mean a user can disable snapshot reference resolution? Do we support the similar switch in other providers?

linglingye001:
> > * Introduce `allowSnapshotRef` to prevent nested snapshot reference resolution and `snapshotClient` to lazily initialize/reuse an App Configuration client for snapshot reads.
> 
> @linglingye001 does it mean a user can disable snapshot reference resolution? Do we support the similar switch in other providers?

No, `allowSnapshotRef` is an internal variable. The k8s provider parses the snapshot references in `processSettings`, fetches the snapshot's settings, `allowSnapshotRef`  false inside a nested snapshot's settings to prevent recursive resolution.

RichardChen820 (approved):

PR #191, opened 2026-08-21
Add validating admission policy

linglingye001:


Makefile
```diff
@@ -119,6 +119,8 @@ build-installer: manifests generate kustomize ## Generate a consolidated YAML wi
 	mkdir -p dist
 	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
 	$(KUSTOMIZE) build config/default > dist/install.yaml
+	printf '\n---\n' >> dist/install.yaml
+	$(KUSTOMIZE) build config/admission >> dist/install.yaml
```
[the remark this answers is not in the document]
linglingye001: The Makefile is for local dev

zhiyuanliang-ms (approved):

config/admission/validatingadmissionpolicy.yaml
jimmyca15, line 1: Why do we have two of these?

linglingye001: This file belongs to `make deploy` installation path, which is for local development/test. Customers use `helm` install and `helm` read the Validation admission policy from the other file.

jimmyca15: Is there some difference between the two such that we can't share?

linglingye001: The file under `admission` is a static manifest consumed by `Kustomize`, while the file under `templates` must be a Helm template so it can conditionally render by Kubernetes version and chart values. `Kustomize` cannot consume Helm syntax, and Helm cannot package a file outside the chart directory. They therefore cannot directly share the same physical file.

